PR #10, opened 2024-02-27
Feature new multinomial model

franciscobmacedo:
Add code to handle new multinomial model. Rename folders for reduced complexity.


tab1tha (approved):
Looks good.

PR #17, opened 2024-03-05
12 add la to sample data

amynickolls:
sample data changed
data container code updated to handle new ingest, keeping YEAR from dataframe rather than metadata

UASC flag added using UASC table - should this sit in a separate function to combined datasets?


ssda903/datacontainer.py
```diff
@@ -71,40 +73,25 @@ def _detect_table_type(
...
-            if metadata.year == year and metadata.table == table_type:
+            if metadata.table == table_type:
                 return self.__datastore.to_dataframe(info)
 
         raise ValueError(
-            f"Could not find table for year {year} and table type {table_type}"
+            f"Could not find table for table type {table_type}"
         )
 
     def get_tables_by_type(
```
franciscobmacedo, line 97: I think you can remove this function? It's not used anywhere I believe. but please do check if it's not used at all in the codebase

franciscobmacedo (changes requested):

ssda903/datacontainer.py
amynickolls: Checked by ctrl+f across codebase and ran sandbox code without error - removed

PR #20, opened 2024-03-08
16 split datacontainer processes, add bootstrap datepicker and implement historic view

franciscobmacedo:
- create read and predict logic in ssda903
- adapt predict view for new code structure and rename it (it was chart view before)
- add bootstrap datepicker - so it looks a bit better!
- add the first version of the historic view, with filters

@amynickolls I left a TODO for you - so you can add the relevant charts in the historic view, according to the designs


dm_regional_app/forms.py
```diff
@@ -1,14 +1,92 @@
...
+        label="Placement Type",
+        required=False,
+        choices=[],
+    )
+
+    def __init__(self, *args, **kwargs):
+        la_choices = kwargs.pop("la_choices")
+        placement_type_choices = kwargs.pop("placement_type_choices")
+
+        super().__init__(*args, **kwargs)
```
amynickolls, line 44: Can we go over this together?

amynickolls (commented):
Looks great

dm_regional_app/forms.py
franciscobmacedo: sure!

MichaelHanksSF (approved):
This all makes sense to me.

PR #21, opened 2024-03-12
created user profile, edited user fixtures+readme

amynickolls:
created user profile with fields show_instruction and la
user creation in fixtures and readme updated to test la extraction from email address


dm_regional_app/migrations/0003_profile_la.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 4.2.10 on 2024-03-12 16:45
+
+from django.db import migrations, models
+
```
franciscobmacedo, line 4: Can you avoid having 2 migration files, like we did in our session yesterday?

dm_regional_app/signals.py
```diff
@@ -0,0 +1,17 @@
...
+from dm_regional_app.models import Profile
+
+User = get_user_model()
+
+
+@receiver(post_save, sender=User)
+def create_user_profile(sender, instance, created, **kwargs):
+    if created:
+        if ".gov.uk" in instance.email:
+            la = instance.email.split("@")[1].split(".")[0]
```
franciscobmacedo, line 14: what if the email is something like "john.gov.uk@gmail.com"? This will not work. Maybe you can check if the email ends with `.gov.uk` with the string `.endswith` method instead of just checking if `.gov.uk` is there

franciscobmacedo (changes requested):

dm_regional_app/migrations/0003_profile_la.py
amynickolls: Is this - only have one migration file for profile, or have one migration overall (init only)?

franciscobmacedo: Yes this it. No, in general you shouldn't change migrations that were  already created before you started doing changes. That would cause an issue in existing databases of this project - for example, the upcoming production database.

PR #30, opened 2024-03-21
Use django storage for files

franciscobmacedo:
Since we will be using a S3 bucket to store data down the line, we need to be able to easily switch between sample data and the production data. With this DataStore implementation, we are able to read the files using Django Storage and, when ready for production, switch to AWS S3 - like it's explained in [this draft issue](https://github.com/orgs/SocialFinanceDigitalLabs/projects/19/views/1?pane=issue&itemId=57230831)


amynickolls (changes requested):
Doesn't look like you've deleted fs_datastore code? Is it used elsewhere?

Might be one to run over on a Wednesday morning but looks good otherwise!

PR #32, opened 2024-03-25
add heroku configuration

franciscobmacedo:
- Added heroku configuration variables and files
- Split settings into different modules, depending on the environment variable `DJANGO_ENV`
- Added `.env.example` and how to use it in the README.md


dm_regional_site/settings/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+DJANGO_ENV = config("DJANGO_ENV", default="production")
+
+if DJANGO_ENV == "development":
+    from .development import *  # NOQA
+
+elif DJANGO_ENV == "github_actions":
+    from .github_actions import *  # NOQA
+
+else:
+    from .production import *  # NOQA
```
christina-moore, line 12: Rather than doing this, we could specify `DJANGO_SETTINGS_MODULE` directly in the environment variables (and then update `manage.py` to point at the env var. Thoughts on this approach? (It would require an adjustment to local setups)

MichaelHanksSF: @cyramic @amynickolls keen to hear your thoughts on this.

README.md
```diff
@@ -16,10 +16,11 @@ This tool is expected to read the data from a remote source (an S3 bucket or sim
...
 1. Create an `.env` file by copying the `.env.example` one and editing the variables as needed.
-2. Install dependencies with `poetry install`;
-3. Launch a poetry shell so the virtual environment is active: `poetry shell`;
-4. Run `python manage.py migrate` to create the databases tables;
-5. Add the pre-commit hook by running `pre-commit install`. This will ensure your code is formatted before you commit something;
+1. Set environment variable `DJANGO_SETTINGS_MODULE=dm_regional_site.settings.development` in relevant places.
+1. Install dependencies with `poetry install`; 
+1. Launch a poetry shell so the virtual environment is active: `poetry shell`; 
+1. Run `python manage.py migrate` to create the databases tables; 
+1. Add the pre-commit hook by running `pre-commit install`. This will ensure your code is formatted before you commit something;
```
tab1tha, line 23: Was it intentional to write all these as ones (1.)?
It is rendered as 1, 2, 3, 4, 5, 6 so I wondered if it was intentional that you wrote them all as 1s.

tab1tha (commented):
The changes look good to me. The part that I'm unsure about is the reason behind them. Was it because we want to ensure that the __init__ file is empty ?

tab1tha (approved):

README.md
christina-moore: Yup, it's a nice little feature of markdown that if you keep them all as ones, it will render them in order. And then it means updating the list doesn't require any changing of the other numbers. The downside is it's a little less clear if folks are viewing the raw files themselves - I'm not married to it so happy if you think it's better the original way 🙂

christina-moore:
> The changes look good to me. The part that I'm unsure about is the reason behind them. Was it because we want to ensure that the **init** file is empty ?

Yes it's my personal preference to keep init files as empty as possible ([this link](https://stackoverflow.com/a/1947834) talks a bit about init setup and I think a clean init helps to promote better practice i.e. importing exactly what is needed to prevent bloat).

The other reason for the changes was that we were using `DJANGO_ENV` alongside `DJANGO_SETTINGS_MODULE` to define the settings used. This way, we only use one environment variable to point the app at the relevant settings

README.md
christina-moore: I've reverted it! Thanks for pointing out the flaws of using `Enter` with this setup

tab1tha:
> > The changes look good to me. The part that I'm unsure about is the reason behind them. Was it because we want to ensure that the **init** file is empty ?
> 
> Yes it's my personal preference to keep init files as empty as possible ([this link](https://stackoverflow.com/a/1947834) talks a bit about init setup and I think a clean init helps to promote better practice i.e. importing exactly what is needed to prevent bloat).
> 
> The other reason for the changes was that we were using `DJANGO_ENV` alongside `DJANGO_SETTINGS_MODULE` to define the settings used. This way, we only use one environment variable to point the app at the relevant settings

Thank you. 
I see your point. 

All looks good to me !

PR #36, opened 2024-04-12
22 create historic view

amynickolls:


dm_regional_app/management/commands/sandbox.py
```diff
@@ -15,13 +15,17 @@
...
-        """datastore = fs_datastore("sample://v1.zip")
+        datastore = fs_datastore("sample://v1.zip")
         config = Config()
         dc = DemandModellingDataContainer(datastore, config)
+        pop = PopulationStats(dc.enriched_view, config)
 
-        
-        print(dc.enriched_view)"""
+        data = dc.enriched_view
+        print(data.loc[data.UASC == True])
```
tab1tha, line 28: flagging in case you forgot to delete the print statements in this file. It's okay if you intentionally left them in.

tab1tha (approved):
Tested locally and everything works. 
All good !

dm_regional_app/management/commands/sandbox.py
amynickolls: Yes these are purely for testing, thanks!

tab1tha: Ah I see !
Let's put a reminder to remove all print statements when this is about to into production since they [can be](https://stackoverflow.com/a/33694403/11044184) security risk.

all good.

PR #43, opened 2024-05-10
23 improve the existing prediction view

amynickolls:
Prediction view brought into line with figma design.

Chart updated.

Historic form updated and added to page.
Predict filter updated.

Some minor text and design changes.

Notes on functionality:
- It doesn't look like the model is working quite as expected, but this is to be solved in another issue.
- Confidence intervals are tiny so you can only view these by zooming in really close


tab1tha (approved):
This looks good.

PR #45, opened 2024-05-30
cleanup: move config to python code

franciscobmacedo:
removed dynamic configuration loader from yaml file (ssda903/fixtures/config/standard-v1.yaml) and moved it into python code.
Now, age brackets, placement categories and costs are all written in python, so everything is detectable by your IDE. 

Also changed all the dataframe values to just have the "label" of the placement category or age bracket - that's the human readable value, which is easy to display in the web but still accessible programmatically, with the enum values.

@amynickolls just be aware that  everytime you need to filter something use it like this:

```python

data.loc[
    data.placement_type_before == PlacementCategories.NOT_IN_CARE.value.label
]
```

Because:
- `PlacementCategories` is an enum
- `PlacementCategories.NOT_IN_CARE` is an enum member
- `PlacementCategories.NOT_IN_CARE.name` is actually "NOT_IN_CARE", the enum member name
- and `PlacementCategories.NOT_IN_CARE.value` is an instance of the dataclass `PlacementCategory` with it's label "Not in care"
- `PlacementCategories.NOT_IN_CARE.value.label` is the name of this `PlacementCategory` instance, so in this case "Not in care"

There's examples in the docstrings of PlacementCategories and AgeBrackets.

I hope now it's more clear!!


amynickolls (approved):
No comments - thanks for the detailed comments in the code!

PR #49, opened 2024-06-28
48 fixes to adjusted view following datacontainer change

amynickolls:
A couple of fixes:
Adjusted page wasn't running post merge, should now be working
Rate adjustments that don't match an initial rate in the prediction will be ignored
Tables on the main adjustment page no longer say 'base', tables on form input page do say base.

### DIFF
--- a/ssda903/multinomial.py
+++ b/ssda903/multinomial.py
@@ -4,7 +4,7 @@ from typing import Any, Iterable, Optional, Union
 
 import numpy as np
 import pandas as pd
-from demand_model.base import BaseModelPredictor, combine_rates
+from demand_model.base import BaseModelPredictor
 from demand_model.multinomial.utils import (
     build_transition_rates_matrix,
     fill_missing_states,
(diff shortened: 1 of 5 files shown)


ssda903/multinomial.py
```diff
@@ -31,6 +31,23 @@ class NextPrediction:
...
+def combine_rates(rate1: pd.Series, rate2: pd.Series) -> pd.Series:
+    """'
+    This has been updated to a multiplication method.
+    Fill value=1 allows any values missing from the adjustment series to remain unchanged in the transition rates
+    Any rates not present in rate1 will not be included in output
+    """
+    rate1, rate2 = rate1.align(rate2, fill_value=1)
+
+    # Create a mask to identify where rate1 is missing but rate2 is not
+    mask = (rate1 != 1) | (rate2 == 1)
```
tab1tha, line 43: This statement will still be true if rate 1 is missing and rate 2 is missing also (`(rate1 != 1) | (rate2 == 1)`) since the first part will make the `or` logic true. Is that okay?
If not, we could consider changing this to an `and` statement so that it is only true if rate 1 is missing and rate 2 is not missing, as the preceding comment describes.
The comment and the code describe slightly different scenarios.

tab1tha (approved):
Apart from clarifying comment, everything else looks good and runs locally.
Good morning !

ssda903/multinomial.py
amynickolls: Thanks - this made me think about this a bit more and I have tightened up the logic!

PR #51, opened 2024-07-17
47 bring costs back into model code

amynickolls:
Added costs.py and code to take the prediction population and create an output of costs.

Base proportions and costs are stored in an enum connected to placementcategories so function retrieves these.

Proportion adjustment logic:

- If a proportion is inputted by the proportion adjustment series, it should remain the same unless related items in the proportion_adjustment series sum to more than 1 (e.g. all fostering must sum to 1)
- Other proportions not in the adjustment series will therefore be scaled down to sum the total to 1
- If proportions in the adjustment series do sum to more than 1 then these will be scaled down to 1 and other related items reduced to 0

Cost adjustment logic:

- Cost adjustments replace enum costs - easy!

There is nothing functional in the django app using this code yet - I suggest using sandbox.py to test out the code. `python manage.py sandbox`

### DIFF
--- a/dm_regional_app/charts.py
+++ b/dm_regional_app/charts.py
@@ -1,8 +1,8 @@
 import pandas as pd
 import plotly.express as px
 import plotly.graph_objects as go
-from demand_model.multinomial.predictor import Prediction
 
+from ssda903.multinomial import Prediction
 from ssda903.population_stats import PopulationStats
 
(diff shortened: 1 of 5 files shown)


tab1tha (approved):
Ran sandbox.py and 
Read the code
This looks good to me.

PR #56, opened 2024-08-09
changes to costs

amynickolls:
Creating a new PR for the changes Michael requested since it's a fair bit of code. Tambe's changes are also in here. The last two bullets were my own changes. To review:

- costs by week, not day
- costs.py now outputs populations. also no longer outputs by bin but by placement category
- charts tidied
- charts show detailed placement type
- logic to deal with proportions summing to <1 added
- DynamicForm no longer accepts negative numbers and logic added to views to warn of this
- summary tables - new table tag written for this to allow subtraction of tables. This means additional code but less processing
- bug fix for inflation on GET request
- clear rate adjustment button on costs page now redirects back to costs

### DIFF
--- a/dm_regional_app/templatetags/table_tags.py
+++ b/dm_regional_app/templatetags/table_tags.py
@@ -81,7 +81,7 @@ def convert_summary_tables_to_html_table(df):
             return f"{value:,.2f}"
         return value
 
-    html = f'<thead><tr><th scope="col" style="padding-top: 8px; padding-bottom: 8px;">{df.index.name}</th>'
+    html = f'<thead><tr><th scope="col" style="padding-top: 8px; padding-bottom: 8px;">Placement</th>'
     for value in df.columns:
         if isinstance(value, str):
             if "base" in value.lower():
--- a/dm_regional_app/templatetags/table_tags.py
+++ b/dm_regional_app/templatetags/table_tags.py
@@ -20,11 +20,10 @@ def convert_data_frame_to_html_table(df):
 
     html = "<thead><tr>"
     for value in df.columns:
-        modified_value = value
         if isinstance(value, str):
             if "base" in value.lower():
-                modified_value = value.lower().replace("base", "").strip().capitalize()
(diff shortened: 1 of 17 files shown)


MichaelHanksSF (changes requested):
Hello @amynickolls,

Thanks for making these changes!

I have made a further commit to amend the code for adjusting proportions. This is because the changes you had made still didn't account for instances where:
- the total < 1
- remaining = 0
I've also reduced the number of ifs involved as all scenarios can be covered by three conditions:
- where total = 1
- where adjusted >= 1 or remaining = 0
- else, which will be total != 1,  adjusted < 1 and remaining > 0

One change that I would like you to make please is to align the naming conventions for the CostForecast class and the convert_population_into_cost() function. I find it confusing to read when e.g. the class considers the costs to be the costs per day, but the function considers costs to be the cost_summary list. Can we align these more closely so that each of them has a name that can't be confused with something else and they map to one another?

@tab1tha Could you please review:
- [x] forms.py
- [x] templates and templatetags changes
- [ ] changes in views.py from line 212 down

Thanks!

dm_regional_app/templatetags/table_tags.py
```diff
@@ -67,3 +66,39 @@ def convert_data_frame_to_html_table_plus_form(df, form, header="Rate multiplica
...
+            if isinstance(value, str):
+                row_html += f'<th scope="row" style="font-size: 15px; padding-top: 8px; padding-bottom: 8px;">{value}</th>'
+            else:
+                formatted_value = format_value(value)
+                row_html += f'<td scope="row" style="font-size: 15px; padding-top: 8px; padding-bottom: 8px;">{formatted_value}</td>'
+        row_html += "</tr>"
+        html += row_html
+
+    html += "</tbody>"
+    return html
```
tab1tha, line 104: Nice !

tab1tha (approved):

MichaelHanksSF (approved):

PR #63, opened 2024-08-23
Datacontainer additional

amynickolls:
- fix start_date definition (lines 231-232) to:
find min(DEC)
if min(DEC) in Apr-Dec (it usually will be); assign start_date as 01/04/{year of DEC}
if min(DEC) in Jan-Mar; assign start_date as 01/04/{year of DEC - 1}

- fix end_date definition (lines 235-236) to:
find max(DEC/DECOM)
if max(DEC/DECOM) in Jan-Mar; assign end_date as 31/03/{year of max DEC/DECOM}
if max(DEC/DECOM) in Apr-Dec; assign end_date as 31/03/{year of max DEC/DECOM +1}

- check add_related_placement_type() (lines 313-338); currently this assigns NOT_IN_CARE where there's a null for the previous or subsequent placement type. 

I probably spent way too long on this! My initial thought was that this doesn't actually matter as we don't allow users to select dates prior to start date or post end date in the app - the only time a not in care marker would be incorrect in this instance is missing data as we don't hold the previous episode. There is no other way of working out an entry/exit from care as far as I can tell. Still went through the motions to prove this theory. The code follows the following logic:
If DECOM before or equal to datacontainer.start_date AND new column currently NAN (this is due to age transitions - there may be 'full placements' prior to start date where they are continuous due to new age placements. This could be adjusted if moving this transformation prior to adding aging placements) - add None for previous placement (Michael you can confirm whether equal to is required in this instance)
If DEC is blank (e.g. placement has no end date) add None for next placement.
Where placements are not continuous, add Not in care for interim placement
For all other Nans, add Not in care.

This did end up leading to a very small change in Entry rates which I've been unable to account for - 
(Original) 10 to 16 Fostering 0.1114 
(New) 10 to 16 Fostering 0.1105 
This is not because of a placement beginning on the start date as the earliest start date is sometime in August. I was running this on the default values if you want to check - so datacontainer start date and end date (the version in this branch) if you want to recreate.

The 'None' values will be automatically ignored as the placement bins are created via the Enum categories rather than the column values.

- add the more granular placement category as a new field for the placements.

### DIFF
--- a/ssda903/config/_costs.py
+++ b/ssda903/config/_costs.py
@@ -15,6 +15,8 @@ class CostDefaults:
 class CostItem:
     label: str
     category: PlacementCategory
+    placement_types: tuple
+    place_provider: tuple
     defaults: CostDefaults
 
     def toJSON(self) -> dict:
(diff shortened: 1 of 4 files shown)


tab1tha (approved):

ssda903/config/_costs.py
```diff
@@ -34,46 +36,97 @@ class Costs(Enum):
...
+        place_provider=("PR4", "PR5"),
         defaults=CostDefaults(cost_per_day=250, proportion=1),
     )
     RESIDENTIAL_IN_HOUSE = CostItem(
         label="Residential (In-house)",
         category=PlacementCategories.RESIDENTIAL.value,
+        placement_types=(
+            "K2",
+            "R1",
+            "P3",
```
MichaelHanksSF, line 79: I added a couple more small placement types to residential

ssda903/config/_costs.py
```diff
@@ -34,46 +36,97 @@ class Costs(Enum):
...
+            "P2",
+        ),
+        place_provider=(),
         defaults=CostDefaults(cost_per_day=1000, proportion=1),
     )
     SECURE_HOME = CostItem(
         label="Secure home",
-        category=PlacementCategories.OTHER.value,
+        category=PlacementCategories.OTHER.value,  # @Michael note - on 903 guidance this is under residential?
+        placement_types=("K1",),
```
MichaelHanksSF, line 114: I had to add a comma after K1 for it to recognise this in the map; otherwise it came out as (K, "") and (1, "")

ssda903/config/_placement_categories.py
```diff
@@ -69,6 +69,8 @@ class PlacementCategories(Enum):
         placement_types=(
             "K2",
             "R1",
+            "P3",
```
MichaelHanksSF, line 72: Added the same additional placement types to residential

ssda903/datacontainer.py
```diff
@@ -225,15 +225,41 @@ def enriched_view(self) -> pd.DataFrame:
...
-        return self.combined_data[["DECOM", "DEC"]].max().max().date()
+        max_dec_decom = self.combined_data[["DECOM", "DEC"]].max().max().date()
+
+        # Extract the month and year from the max_dec_decom date
+        max_dec_decom_month = max_dec_decom.month
+        max_dec_decom_year = max_dec_decom.year
+
+        # Determine the end_date based on the month of max_dec_decom
+        if 4 <= max_dec_decom_month <= 12:  # April to December
+            end_date = date(max_dec_decom_year + 1, 3, 31)
```
MichaelHanksSF, line 258: This needs to be +1 for end date

ssda903/datacontainer.py
```diff
@@ -318,23 +344,32 @@ def _add_related_placement_type(
 
         WARNING: This method modifies the dataframe in place.
         """
-
         combined = combined.sort_values(["CHILD", "DECOM", "DEC"], na_position="first")
 
+        # Group by child and set next/preceding placement based on offset
```
MichaelHanksSF, line 349: I undid quite a bit of this - sorry! I don't think my directions were particularly clear.

I think that the preceding episode should be Not in Care if there is no preceding episode for a child, even if that is just because we don't know what that episode was. It's a downside of the dataset and there's nothing much we can do about it.

The addition of the DEC.isna() change is good, but as you say, it may not have any impact in reality.

I think the rest of the original code was fine and a little more compact.

MichaelHanksSF (approved):

PR #79, opened 2024-09-11
refactor: centralise messages on base template

christina-moore:
Centralising the messages onto the base template and mapping `error` as I'll need that one for the upload data page


amynickolls (approved):
Thanks for this - so much better!

PR #35, opened 2024-03-28
configure AWS s3 bucket for production

franciscobmacedo:
S3 is configured for storage (both for upload of files and reading of files)

The infrastructure is terraformed - a remote backend is used so state files are stored in a manually created S3 bucket (demand-model-terraform-state)

PR #80, opened 2024-09-16
Add view for staff users to upload files

christina-moore:
Notes:
- I tested locally with an episodes file of ~100k lines and a header file of ~20k lines. This request took ~50 seconds to complete which would timeout the webpage in production. If we think this is a size of data the tool should be able to handle, we might need to consider either looking at the validation (full prediction), optimising the prediction itself, or doing the validation asynchronously. My first recommendation would be optimising the prediction itself as it occurs in other web requests as well.
- I think we should add tests for the prediction function itself, but that can be a separate PR
- This PR relies on https://github.com/SocialFinanceDigitalLabs/cs-demand-model-regional/pull/35 to test S3 storage
- Once SSO is implemented, this view should only be accessible by SSO-authenticated users

image
image


amynickolls (approved):

amynickolls:
> * I tested locally with an episodes file of ~100k lines and a header file of ~20k lines. This request took ~50 seconds to complete which would timeout the webpage in production. If we think this is a size of data the tool should be able to handle, we might need to consider either looking at the validation (full prediction), optimising the prediction itself, or doing the validation asynchronously. My first recommendation would be optimising the prediction itself as it occurs in other web requests as well.

On this - Michael will have a better view than me on the population we'll be expecting from a Region. Looked after children are less than 100k in the UK but rows are episodes rather than children.

PR #83, opened 2024-09-16
Placement proportion output

amynickolls:
- Updated populationstats to output historic proportions
- Updated convert_population_to_cost to take historic proportions rather than pulling defaults from enum
- Costs enum default proportions deleted
- Finished logic on cost page to include historic placement proportions showing adjusted options and update table to show historic and current proportions
- Clear adjusted proportions button added
- Update historic costs and numbers part of costs and numbers so this runs on detailed historic population rather than forecast output

image


dm_regional_app/charts.py
```diff
@@ -33,7 +33,7 @@ def year_one_costs(df: CostForecast):
     return total_sum
 
 
-def area_chart_cost(historic_data: CostForecast, prediction: CostForecast):
+def area_chart_cost(df_historic, prediction: CostForecast):
```
christina-moore, line 36: Curious if it's worth adding a type to the new parameters in the functions of this file?

dm_regional_app/templates/dm_regional_app/views/costs.html
```diff
@@ -243,6 +244,9 @@ <h2 class="accordion-header" id="headingThree">
                         </table>
                     </div>
                     <div class="d-flex bd-highlight mb-3">
+                      {% if proportions.columns|length == 4 %}
+                      <div class="p-2 bd-highlight"><a class="btn btn-warning" href="clear_proportions?next_url_name=costs">Clear all proportion adjustments</a></div>
```
christina-moore, line 248: A nitpick, but I think putting the <a> tags on a newline could improve readability
```suggestion
                      <div class="p-2 bd-highlight">
                          <a class="btn btn-warning" href="clear_proportions?next_url_name=costs">Clear all proportion adjustments</a>
                      </div>
```

ssda903/utils.py
```diff
@@ -0,0 +1,12 @@
...
+
+def get_cost_items_for_category(category_label: str):
+    """
+    Takes a placement category label and returns related enum costs in a list
+    """
+    cost_items = []
+    for cost in Costs:
+        if cost.value.category.label == category_label:
+            cost_items.append(cost.value)
+    return cost_items
```
christina-moore, line 12: You could use a list comprehension here (dealer's choice on readability)
```suggestion
    return [cost.value for cost in Costs if cost.value.category.label == category_label]
```

ssda903/utils.py
```diff
@@ -0,0 +1,12 @@
+from ssda903.config import Costs
+
+
+def get_cost_items_for_category(category_label: str):
```
christina-moore, line 4: I do think it makes sense for this to live on the Costs enum as a class method

christina-moore (approved):

PR #84, opened 2024-09-17
59 add graph lines showing model states

WillLP-code:
#closes 59

- Adds traces to historical and forecast charts breaking down forecast and historical data (where appropriate) into care type bucket.
- Uses a colour blind accessible colour palette (may not translate to grayscale but at that point we ought to add more than colour to differentiate between traces like triangles, crosses, squares etc).
- Allows highlighting/filtering traces using plotly's built in functionality.
- Splits most of the work into functions in the utils.py as they're used across both plotting functions.
- A bunch of these sub-functions use loops per care type, if there's a more efficient way I'm happy to change how its done.
- If anything isn't done in a preferred way/style etc. let me know!

### DIFF
--- a/dm_regional_app/charts.py
+++ b/dm_regional_app/charts.py
@@ -177,19 +177,19 @@ def prediction_chart(historic_data: PopulationStats, prediction: Prediction, **k
 
     df.columns = ["from", "date", "forecast"]
     # Organises forecast data into dict of dfs by care type bucket
-    forecast_care_by_type_dfs = care_type_organiser(df)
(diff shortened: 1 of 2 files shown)


amynickolls (changes requested):

amynickolls (changes requested):
Apologies if I missed this first time round but I'm getting an error on the historic page - `care_type_organiser() missing 2 required positional arguments: 'data_type' and 'input_col'`

Apart from that, changes look good!

amynickolls (approved):

PR #85, opened 2024-09-17
new placement_starts_chart() and associated chgs

robjharrison:
spelling typo fix on readme

placements counts chart on historic page. includes additional start/end date vars in views to enable possible restriction/changes in timeframe of placements chart. Avg placement duration shows only in plot hover.


dm_regional_app/charts.py
```diff
@@ -282,6 +283,88 @@ def historic_chart(data: PopulationStats):
...
+    # calculate placement duration (end_age - age)
+    df_filtered["placement_duration"] = df_filtered["end_age"] - df_filtered["age"]
+
+    df_entrants = (
+        df_filtered.groupby(["DECOM", "placement_type"])
+        .agg(
+            count=("CHILD", "size"), avg_duration=("placement_duration", "mean")
+        )  # calculate count and avg duration
+        .reset_index()
+    )
```
amynickolls, line 305:
```suggestion
    df_entrants = (
        df_filtered.groupby(["DECOM", "placement_type"])
        .agg(
            count=("CHILD", "size"), avg_duration=("placement_duration", "mean")
        )  # calculate count and avg duration
        .reset_index()
    )
        # Create a range of all months to ensure we have complete months
    all_months = pd.date_range(start=start_date, end=end_date, freq='M')

    # Group by 'placement_type' and aggregate
    df_entrants = (
        df_filtered.groupby(["placement_type", pd.Grouper(key="DECOM", freq="M")])  # Group by month
        .agg(
            count=("CHILD", "size"), 
            avg_duration=("placement_duration", "mean")
        )
        .reset_index()  # Reset index after grouping
    )

    # Reindex the result to include all months and fill missing with 0
    df_entrants = df_entrants.set_index(['DECOM', 'placement_type']).reindex(
        pd.MultiIndex.from_product([all_months, df_entrants['placement_type'].unique()], names=['DECOM', 'placement_type']),
        fill_value=0
    ).reset_index()
```
Need to fill missing months with 0 - this is a fairly convoluted way of doing it - feel free to make it simpler

dm_regional_app/templates/dm_regional_app/views/historic.html
```diff
@@ -14,18 +14,29 @@ <h1>Viewing the SSDA903 data</h1>
...
+        </div>
     </div>
-    <div class="col">
-        <h1>Total # of exiting care</h1>
-        <p><h1 class="display-2">{{ exiting_care_count }}</h1></p>
+    <br>
+
+    <!-- Historic Chart - ref: was historic_chart -->
+    <div class="alert alert-primary" role="alert">
+        <h3>Historic Data</h3>
```
amynickolls, line 30: Remove title - title is within chart

dm_regional_app/templates/dm_regional_app/views/historic.html
```diff
@@ -14,18 +14,29 @@ <h1>Viewing the SSDA903 data</h1>
...
+        {{ historic_chart|safe }}
+    </div>
+
+    <!-- Placement Starts Chart -->
+    <div class="alert alert-primary" role="alert">
+        <h3>Placement Starts</h3>
```
amynickolls, line 36: Remove title - title is within chart

dm_regional_app/views.py
```diff
@@ -1159,7 +1177,8 @@ def historic_data(request):
 
         stats = PopulationStats(data)
 
-        chart = historic_chart(stats)
+        chart = historic_chart(stats)  # consider name revision
```
amynickolls, line 1180: historic_population_chart?

amynickolls: On this - perhaps update the title of historic data chart to 'Historic population' (within charts.py)

dm_regional_app/charts.py
```diff
@@ -282,6 +283,88 @@ def historic_chart(data: PopulationStats):
...
+
+    fig.update_traces(line=dict(width=1), marker=dict(size=4))
+
+    max_count = df_entrants["true_count"].max()
+    fig.update_yaxes(
+        title_text="Placements",
+        tickvals=list(range(0, int(max_count) + 10, 10)),
+        ticktext=[str(i) for i in range(0, int(max_count) + 10, 10)],
+    )
+
```
amynickolls, line 363: Need to set axes - x to 0, y to match 1st and last data point (currently some white space either end, I think due to the way years are labelled)

amynickolls (dismissed):

dm_regional_app/charts.py
robjharrison: Ah yes, sorry i forgot about those blank ends. Meant to sort that.

dm_regional_app/charts.py
```diff
@@ -282,6 +283,88 @@ def historic_chart(data: PopulationStats):
     return fig_html
 
 
+def placement_starts_chart(data: PopulationStats, start_date: str, end_date: str):
+    data.df["DECOM"] = pd.to_datetime(data.df["DECOM"])
+    start_date, end_date = pd.to_datetime([start_date, end_date])
+    data.df["DECOM"] = data.df["DECOM"].dt.to_period("M").dt.to_timestamp()
```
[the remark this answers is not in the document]
robjharrison: Not sure how you're getting this to work without this, I've not been able. I'm leaving it in for now as without i've only type incompatibilities on dt.

robjharrison:
Amy. I've made the changes, excepting the date conversions, see comment. Apologies for the back and forth on this.

dm_regional_app/charts.py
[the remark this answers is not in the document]
amynickolls: No problem, it might be a codespaces inconsistency as Will mentioned

amynickolls:
No problem on the back and forth - afraid I have a couple more requests of you on this!

Thanks for resetting the axes, but could we get a bit of space at the top axis of the chart? As in the historic data chart.

image

I've also just merged in Rob's PR on getting placements into the historic graph. Would you mind merging main into this branch and checking that the colours are consistent for placements across the two graphs? Thanks!

dm_regional_app/views.py
```diff
@@ -1135,8 +1136,21 @@ def historic_data(request):
                 # update reference start and end
 
                 data = apply_filters(datacontainer.enriched_view, form.cleaned_data)
+
+                # # Get start_date and end_date from form.cleaned_data
+                # start_date = form.cleaned_data["start_date"]
+                # end_date = form.cleaned_data["end_date"]
```
amynickolls, line 1142: Missed these first time I think - left in comments/outdated comments

dm_regional_app/views.py
```diff
@@ -1150,6 +1164,10 @@ def historic_data(request):
             )
             data = apply_filters(datacontainer.enriched_view, form.initial)
 
+            # get datacontainer dates
+            start_date = form.initial.get("start_date", datacontainer.start_date)
+            end_date = form.initial.get("end_date", datacontainer.end_date)
```
amynickolls, line 1169: There aren't any dates in the form, so just grab these from datacontainer. You could probably also move these out of the if/else since these should stay the same. I think even where there is a filtered population to cover less time (e.g. no children UASC of a certain age for specific years) we would still want to show in the table that the data was none for that period rather than reducing the period shown in the chart.

amynickolls (commented):
Apologies as I think I missed these two items re start/end dates and comments in the first review!

dm_regional_app/views.py
robjharrison: Agreed, have extended this, to the additional removal/refactor of duplicated start/end date retrieval within the IF block. start_date and end_date now only extracted once after datacontainer loaded.

dm_regional_app/views.py
robjharrison: thanks, edited charts.py

dm_regional_app/charts.py
```diff
@@ -276,12 +277,126 @@ def historic_chart(data: PopulationStats):
...
+        .reindex(
+            pd.MultiIndex.from_product(
+                [all_months, df_entrants["placement_type"].unique()],
+                names=["DECOM", "placement_type"],
+            ),
+            fill_value=0,
+        )
+        .reset_index()
+    )
+
```
MichaelHanksSF, line 321: Bit more commentary on lines 310-320 would be appreciated

dm_regional_app/charts.py
```diff
@@ -276,12 +277,126 @@ def historic_chart(data: PopulationStats):
...
+        markers=True,
+        line_shape="spline",
+        labels={  # hover dict
+            "true_count": "Placements",
+            "DECOM": "Date",
+            "placement_type": "Placement Type",
+            "avg_duration_weeks": "Avg Duration (wks)",
+        },
+        # title="Placement starts per month by Placement Type",
+        hover_data={"true_count": True, "avg_duration_weeks": True},
```
MichaelHanksSF, line 345: As Amy suggested, the commented out line can be removed.

dm_regional_app/charts.py
```diff
@@ -276,12 +277,126 @@ def historic_chart(data: PopulationStats):
...
+
+    quarterly_months = df_entrants[df_entrants["DECOM"].dt.month.isin([1, 4, 7, 10])]
+    fig.update_layout(
+        xaxis=dict(
+            tickvals=quarterly_months["DECOM"],
+            ticktext=[f"{d:%b}" for d in quarterly_months["DECOM"]],
+            tickangle=45,
+            tickmode="array",
+            title=dict(text="Date", standoff=40),
+        ),
```
MichaelHanksSF, line 356: I think the title "Date" isn't too informative, when the axis is showing months - can we switch to month?

dm_regional_app/charts.py
```diff
@@ -276,12 +277,126 @@ def historic_chart(data: PopulationStats):
...
+            ),
+            fill_value=0,
+        )
+        .reset_index()
+    )
+
+    # convert avg_duration yrs to more sensical wks (using 52.14 wks pa)
+    df_entrants["avg_duration_weeks"] = (df_entrants["avg_duration"] * 52.14).round(2)
+
+    df_entrants["true_count"] = df_entrants["count"]
```
MichaelHanksSF, line 325: Is there a need for this true count column?

dm_regional_app/charts.py
```diff
@@ -276,12 +277,126 @@ def historic_chart(data: PopulationStats):
...
+    fig.update_layout(
+        xaxis=dict(
+            tickvals=quarterly_months["DECOM"],
+            ticktext=[f"{d:%b}" for d in quarterly_months["DECOM"]],
+            tickangle=45,
+            tickmode="array",
+            title=dict(text="Date", standoff=40),
+        ),
+        title="Placement starts per month by Placement Type",
+        hovermode="x",
```
MichaelHanksSF, line 358: I prefer to not have the hovermode = "x". I think it's more legible when you're only seeing the hover for the line you're looking at.

dm_regional_app/charts.py
MichaelHanksSF: Also, this change has the unfortunate effect of switching from showing the year month in the hover to showing the last date of the month for the non quarterly months and just Jan, Apr, Jul for the quarterly months.

dm_regional_app/charts.py
```diff
@@ -276,12 +277,126 @@ def historic_chart(data: PopulationStats):
...
+            x=f"{year}-07-01",
+            y=-0.18,
+            text=str(year),
+            showarrow=False,
+            xref="x",
+            yref="paper",
+            align="center",
+        )
+
+    fig.update_traces(line=dict(width=1), marker=dict(size=4))
```
MichaelHanksSF, line 373: Is it necessary to call this as a separate function, rather than setting the width and the marker size in the original line function?

dm_regional_app/charts.py
```diff
@@ -276,12 +277,126 @@ def historic_chart(data: PopulationStats):
...
+
+    # min and max date vals for x
+    fig.update_xaxes(
+        title_text="Date",
+        range=[
+            df_entrants["DECOM"].min(),
+            df_entrants["DECOM"].max(),
+        ],  # Use min and max date values
+    )
```
MichaelHanksSF, line 382: Again, fairly sure you can set these in the original graph call?

dm_regional_app/charts.py
```diff
@@ -276,12 +277,126 @@ def historic_chart(data: PopulationStats):
...
+        title_text="Placements",
+        range=[
+            min_count,
+            max_count * 1.1,
+        ],  # raise y-axis to give top end data visual space
+        tickvals=list(
+            range(int(min_count), int(max_count) + 10, 10)
+        ),  # generate tick marks
+        ticktext=[str(i) for i in range(int(min_count), int(max_count) + 10, 10)],
+    )
```
MichaelHanksSF, line 395: And again - original graph call?

Happy for you to split that call out on the page to add more commentary, but it seems unnecessary to be creating and re-creating this graph so many times.

dm_regional_app/charts.py
```diff
@@ -276,12 +277,126 @@ def historic_chart(data: PopulationStats):
...
 
+def placement_starts_chart(data: PopulationStats, start_date: str, end_date: str):
+    data.df["DECOM"] = pd.to_datetime(data.df["DECOM"])
+    start_date, end_date = pd.to_datetime([start_date, end_date])
+    data.df["DECOM"] = data.df["DECOM"].dt.to_period("M").dt.to_timestamp()
+
+    # filter time-frame to form dates
+    df_filtered = data.df[
+        (data.df["DECOM"] >= start_date) & (data.df["DECOM"] <= end_date)
+    ]
```
MichaelHanksSF, line 294: Unfortunately, we have a bigger problem here. The data has been enriched to include new "placements" for when children change age bin categories. This is so that ageing transition rates can be explicitly calculated along with other types of transition. But here, those are not legitimate placement starts, so we need to filter them back out again. You're looking for episodes where the only difference between the previous bin and the current bin is the age bucket.

dm_regional_app/charts.py
```diff
@@ -276,12 +277,126 @@ def historic_chart(data: PopulationStats):
...
+    # min and max vals for y
+    min_count = df_entrants["true_count"].min()
+    max_count = df_entrants["true_count"].max()
+
+    # visualise
+    fig = px.line(
+        df_entrants,
+        y="true_count",
+        x="DECOM",
+        color="placement_type",
```
MichaelHanksSF, line 335: I would also like alignment with the other graphs in terms of colours. This is achieved in the other graphs by using the util functions add_traces. If you don't use this function (which I think you could if you wanted to from the data you have here), could you align the colours of the lines and add a line for total please?

MichaelHanksSF (changes requested):
Hi Rob, thanks for this work. I like some of the additional touches, like the hover (though see comment below!). 

I have merged this branch with main to add the work that has been done on other graphs, as I thought it might be helpful. You don't need to use this, but as in comments below, we do need to align the colour scheme for this graph with the other graphs in the app.

dm_regional_app/charts.py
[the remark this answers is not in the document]
robjharrison: Have removed 287 (data.df["DECOM"] = pd.to_datetime(data.df["DECOM"])), but 289 now still required with subsequent later changes.

dm_regional_app/charts.py
robjharrison: there is not, thanks. This was a run-over from some previously needed chart related coercions.

dm_regional_app/charts.py
robjharrison: application of add_trace is problematic, despite the data being apparently aligned. Ref: This function wasn't available at the time of initial dev on this ticket. But formatting is now aligned with that and new total line added.

dm_regional_app/charts.py
robjharrison: Ah yes, i missed that, month is new label now.  I'm grateful for you spotting the other date anomaly(!).. I need to return to that, not yet resolved.

dm_regional_app/charts.py
robjharrison: Have now reset this to align with the historic data chart, showing single trace vals.

dm_regional_app/charts.py
robjharrison: Yes this shouldn't be the case. Will refactor.

dm_regional_app/charts.py
robjharrison: Yes this shouldn't be the case. Will refactor. 
(Some of this was a run-over from a point where getting some of the needed data points to show had been problematic and forced a less than ideal approach)

MichaelHanksSF (changes requested):
Hi Rob, almost there for sure!

I have a couple of very small changes:
- there's a test comment in line 314 - can we remove this please?
- because of the different tick marks we get an inconsistent view in the hover. Can we fix this by adding as a setting to the first fig.add_trace() call lines 351-360:
hovertemplate="%{x|%b %Y}, %{y}"

This hasn't addressed the issue of the age transitions, but we can address that in a separate PR.

Before you can request the merge, you'll also need to manage the conflicts with main. If this is an issue, can you make your commits for the above changes and I'll do the merge.

Thanks,

Michael

MichaelHanksSF (approved):

PR #86, opened 2024-09-24
78 adjusted forecast graph enhancement

WillLP-code:
#closes 78

Allows the user to see the updated and original forecast on the /adjusted pages, where the original only showed the updated forecast.

### DIFF
--- a/dm_regional_app/views.py
+++ b/dm_regional_app/views.py
@@ -973,11 +973,42 @@ def adjusted(request):
                 rate_adjustment=session_scenario.adjusted_rates,
                 number_adjustment=session_scenario.adjusted_numbers,
             )
+            if session_scenario.adjusted_numbers is not None:
+                # if previous rate adjustments have been made, update old series with new adjustments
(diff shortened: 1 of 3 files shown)


dm_regional_app/views.py
```diff
@@ -966,24 +966,45 @@ def adjusted(request):
...
-                number_adjustment=session_scenario.adjusted_numbers,
+            original_prediction = predict(
+                data=historic_data, **session_scenario.prediction_parameters
             )
 
-            # build chart
-            chart = prediction_chart(
-                stats, prediction, **session_scenario.prediction_parameters
-            )
+            if session_scenario.adjusted_numbers is not None:
```
amynickolls, line 973:
```suggestion
            if session_scenario.adjusted_numbers is not None or session_scenario.adjusted_rates is not None:
```

amynickolls (changes requested):

amynickolls (approved):

PR #87, opened 2024-09-24
feat: restrict admin view for staff users

christina-moore:
Only superusers can update the flags indicating if users are staff or superusers - this is to prevent staff users from mistakenly giving regular users the power to upload data
 
image


tab1tha (approved):
Looks good to me.

PR #88, opened 2024-09-24
74 update transition rates table in respect of age transitions

WillLP-code:
#closes 74

Removes aging out as a transition for transition rate table on /adjusted when looking at the tables.


dm_regional_app/utils.py
```diff
@@ -86,3 +85,59 @@ def number_format(value):
...
+    """
+    Used to remove age transitions from transitions rate table
+    """
+    age_up_16 = df["from"].str.contains("10 to 16") & df["to"].str.contains("16 to 18+")
+    age_up_10 = df["from"].str.contains("5 to 10") & df["to"].str.contains("10 to 16")
+
+    df = df[~(age_up_16 | age_up_10)]
+
+    df
+    return df
```
amynickolls, line 100: This will break if a user reconfigures the age brackets. I think best approach is to extract the age brackets from the string and exclude where these aren't equal in the two columns.

dm_regional_app/utils.py
```diff
@@ -86,3 +85,59 @@ def number_format(value):
...
+
+    Transition rate tables need to be sorted on 'From' and 'To' in order
+    to maintain sorting across tables and columns.
+    """
+    # Making age and placement columns for sorting
+    df["split"] = df[bin_col].str.split(" ")
+    df["first_age"] = df["split"].str[0].astype("int")
+    df["starting_place"] = df["split"].str[-1]
+
+    # Sorts entry and exit ratre tables by age then placement
```
amynickolls, line 116: Typo! (Exit rate)

amynickolls (changes requested):

amynickolls (approved):

PR #93, opened 2024-09-30
feat: implement sso

christina-moore:
I've overridden the django-allauth templates where necessary. Happy to continue to modify them to suit desired UI.

image

image

image

image

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -6,3 +6,6 @@ ALLOWED_HOSTS=localhost,127.0.0.1
 
 DATABASE_URL=postgres://postgres:postgres@127.0.0.1:5432/demandmodel
 
+MICROSOFT_CLIENT_ID=microsoft_client_id
+MICROSOFT_CLIENT_SECRET=microsoft_client_secret
+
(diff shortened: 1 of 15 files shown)


tab1tha (approved):
Tested and the code works on my computer too. 
Something small to flag is the use of sign in/out vs log in/out. A user who clicks the `login` button is directed to a page asking them to `sign in` (... with Microsoft) and user who clicks the `logout` button is asked if they're sure that they want to `sign out` .
Maybe we could use the same words across the user journey.

christina-moore:
> Tested and the code works on my computer too. Something small to flag is the use of sign in/out vs log in/out. A user who clicks the `login` button is directed to a page asking them to `sign in` (... with Microsoft) and user who clicks the `logout` button is asked if they're sure that they want to `sign out` . Maybe we could use the same words across the user journey.

Nice spot! I have updated the "Sign Ins" and "Sign Outs" to logins and logouts 👍

amynickolls (approved):
Ran locally and all working.

PR #94, opened 2024-10-02
issue with adjusted rates breaking projected costs page identified

### LINKED ISSUE #91: Error: Costs page
ValueError at /costs
Columns must be same length as key

Error coming from exit rate table: to recreate error, save adjustments to transition rates ONLY and try costs page.

Probably just needs additional breakpoint and transition rate table will need same logic applied for when exit rates but no transition rates - currently this doesn't error but shows empty table which we don't want.

WillLP-code:
closes #91 

If changes are made to transition rates, when the exit rate table is calculated for the projected costs page, this gives an error where two columns are meant to be made from a .str.split() and no split can occur. This error occurs when there are no rows of data to split (after removing all data except rows where children are leaving care).

I've temporarily popped a try/except block in there as this works as a fix, but I'm not sure if this is an actual long term solution. Also, I wonder if we might need to account for this in other instances if there's a chance we might be trying to split where there is no data to split. If we find a long term solution, it might make sense to handle this in another function and use it each time we do thsi split?

### DIFF
--- a/dm_regional_app/charts.py
+++ b/dm_regional_app/charts.py
@@ -566,10 +566,11 @@ def transition_rate_changes(base, adjusted):
 
     df = df[df["base"] != df["adjusted"]]
 
+    df = transition_rate_table(df)
+
     if df.empty:
         return None
     else:
-        df = transition_rate_table(df)
         df = df[["From", "To", "base", "adjusted"]]
         df.columns = ["From", "To", "Base transition rate", "Adjusted transition rate"]
         df = df.round(4)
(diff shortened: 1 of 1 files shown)


amynickolls:
Just a reminder on this one - we want to add logic so that users won't be displayed an empty dataframe as in image below. I beleive the logic for this is already in the view/template, but looks like the if statement in the charts.py functions (for all rates tables) is in the wrong place to return None. Could you fix that on this branch please?

image

WillLP-code:
> 0aafec5
 Hi @amynickolls 

I think this is working as expected now, and I think I've understood what was asked. I was confused as there is now no exit rate table button for me to check if it has an empty dataframe in, but of course there is not, if it gets none, it doesn't make the button

amynickolls (approved):
Yes, looks great!

PR #96, opened 2024-10-03
75 add option to sum in editing rates

amynickolls:
Users can now choose between multiplying or summing rates when editing these (exit rates, entry rate, and transition rates)

DynamicRateForm 
Takes a dataframe, from which it instantiates multiply and add fields
Outputs a dataframe from the save function with multiply and add columns, leaves out rows where both values are empty
Negative numbers not allowed in multiply fields
One one of multiply or add can be used in each row
Displays initial data where previous values have been saved

image

Updates to JSONencoder/decoder to handle dataframes (both multiindex and single index)

Updates to combine_rates in Multinomial to handle new logic: rate * rate_change or + rate_change dependant on column. No longer left joins so users can edit rates at 0 in original prediction.

Updates to templatetags to show field errors in forms.

Unit tests added for DynamicRateForm


dm_regional_app/templates/dm_regional_app/views/entry_rates.html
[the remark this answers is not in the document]
tab1tha: Will do.

dm_regional_app/utils.py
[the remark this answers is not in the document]
amynickolls: This is the JSON encoder/decoder - this worked for converting the rate adjustment series into JSON format to store in the database as part of the scenario. Since the updated method now returns a dataframe it's now been updated to convert multi/single index dataframes too. @christina-moore could check this?

dm_regional_app/views.py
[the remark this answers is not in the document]
amynickolls: Combine first on paper should have worked but when testing it wasn't prioritising 0's coming in from the new data. I don't understand your comment re: old data being used - we want to keep old data where there isn't new data coming in.

ssda903/multinomial.py
```diff
@@ -31,21 +31,44 @@ class NextPrediction:
...
+    """
+    # Ensure rate_adjustment has the required columns
+    if not {"multiply_value", "add_value"}.issubset(rate_adjustment.columns):
+        raise ValueError(
+            "Adjusted rate dataframe must have 'multiply_value' and 'add_value' columns."
+        )
+
+    # Align original_rate and rate_adjustment DataFrame with an outer join to retain all indices
+    original_rate, rate_adjustment = original_rate.align(rate_adjustment, join="outer")
+
```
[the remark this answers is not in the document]
amynickolls: Sure

ssda903/multinomial.py
```diff
@@ -31,21 +31,44 @@ class NextPrediction:
...
+    - Adds 'add_value' to original_rate where provided.
+    - Ensures the result of addition is not less than 0.
+    - Excludes cases where original_rate is missing.
+    """
+    # Ensure rate_adjustment has the required columns
+    if not {"multiply_value", "add_value"}.issubset(rate_adjustment.columns):
+        raise ValueError(
+            "Adjusted rate dataframe must have 'multiply_value' and 'add_value' columns."
+        )
+
```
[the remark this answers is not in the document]
amynickolls: I don't think there would be any experience for the user - the site would crash. Only reason for it to happen is developer changing the flow - useful for debugging. Perhaps @cyramic could check this? I added in the wake of his logging dev session but might have made some wrong assumptions!

dm_regional_app/forms.py
```diff
@@ -325,3 +325,124 @@ class DataSourceUploadForm(forms.Form):
...
+            if initial_data is not None:
+                try:
+                    initial_multiply = initial_data.loc[index, "multiply_value"]
+                except (KeyError, IndexError, ValueError):
+                    pass
+
+                try:
+                    initial_add = initial_data.loc[index, "add_value"]
+                except (KeyError, IndexError, ValueError):
+                    pass
```
christina-moore, line 359: It's the exact same functionality, but I wonder if this would be a bit more intuitive for devs new to the code
```suggestion
            # Set initial values if available
            if initial_data is not None:
                try:
                    initial_multiply = initial_data.loc[index, "multiply_value"]
                except (KeyError, IndexError, ValueError):
                    initial_multiply = None

                try:
                    initial_add = initial_data.loc[index, "add_value"]
                except (KeyError, IndexError, ValueError):
                    initial_add = None
```

dm_regional_app/forms.py
```diff
@@ -325,3 +325,124 @@ class DataSourceUploadForm(forms.Form):
...
+                )
+
+        for index in self.dataframe.index:
+            multiply_field_name = f"multiply_{index}"
+            multiply_value = cleaned_data.get(multiply_field_name)
+            if multiply_value is not None and multiply_value < 0:
+                self.add_error(
+                    multiply_field_name,
+                    "You cannot multiply a rate by a negative value",
+                )
```
christina-moore, line 404: Can this be done in the loop above?

dm_regional_app/forms.py
```diff
@@ -325,3 +325,124 @@ class DataSourceUploadForm(forms.Form):
...
+        multiply_values = []
+        add_values = []
+
+        # Loop through the form fields and collect the input values
+        for index in self.dataframe.index:
+            multiply_value = self.cleaned_data.get(f"multiply_{index}", None)
+            add_value = self.cleaned_data.get(f"add_{index}", None)
+
+            if multiply_value is not None or add_value is not None:
+                # multiply_value = np.nan if multiply_value is None else multiply_value
```
christina-moore, line 422: Should these lines be removed?

dm_regional_app/forms.py
```diff
@@ -325,3 +325,124 @@ class DataSourceUploadForm(forms.Form):
...
+        # Create a DataFrame with the collected input values
+        data = pd.DataFrame(
+            {
+                "transition": transitions,
+                "multiply_value": multiply_values,
+                "add_value": add_values,
+            }
+        )
+
+        # Convert the transition column to a MultiIndex if it contains tuples
```
christina-moore, line 437: In what cases will this happen?

dm_regional_app/templatetags/table_tags.py
```diff
@@ -102,3 +124,73 @@ def format_value(value):
 
     html += "</tbody>"
     return html
+
+
+@register.filter
+def convert_df_and_dynamicrateform_to_table(df, form):
```
christina-moore, line 130: I'm curious about these functions - is the end goal to create tables from the dynamic form?

dm_regional_app/tests/test_dynamicrateform.py
christina-moore, line 1: Nice test suite 🤓

dm_regional_app/tests/test_dynamicrateform.py
```diff
@@ -0,0 +1,122 @@
+import numpy as np
+import pandas as pd
+from django.test import TestCase
+
+from dm_regional_app.forms import DynamicRateForm
+
+
+class DynamicRateFormTest(TestCase):
+    def setUp(self):
```
christina-moore, line 9: `setUp` runs once per test whereas `setUpTestData` (a class method) runs once per TestCase class so can be used for setup that's shared

dm_regional_app/tests/test_dynamicrateform.py
```diff
@@ -0,0 +1,122 @@
...
+            form_data, dataframe=self.dataframe, initial_data=self.initial_data
+        )
+
+        # Check if the form is valid
+        self.assertTrue(form.is_valid())
+
+        # After cleaning, check if cleaned_data is as expected
+        cleaned_data = form.clean()
+        self.assertEqual(cleaned_data["multiply_('A', 'X')"], 1.3)
+        self.assertTrue(cleaned_data["multiply_('B', 'Y')"] == None)
```
christina-moore, line 51: Could use [`assertIsNone`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIsNone) here

dm_regional_app/tests/test_dynamicrateform.py
```diff
@@ -0,0 +1,122 @@
...
+        # Test form submission with invalid data (negative multiply numbers)
+        form_data = {
+            "multiply_('A', 'X')": -1.3,
+            "add_('A', 'X')": "",
+            "multiply_('B', 'Y')": "",
+            "add_('B', 'Y')": 0.9,
+            "multiply_('C', 'Z')": "",
+            "add_('C', 'Z')": "",
+        }
+
```
christina-moore, line 85: Nice one on testing multiple validation failures! Might be good to separate out this into another test and then you can give the two tests more specific names based on what they're testing

dm_regional_app/forms.py
christina-moore: I think the test case sort of answered my question here (yay tests!), but I think I'd still like to chat through the scenarios to understand how the form fields and therefore the dataframe columns could look

dm_regional_app/templatetags/table_tags.py
amynickolls: Yes, we end up with the form input integrated into the table - means we can make it look like a multi-index and get the original data in there too 

image

dm_regional_app/utils.py
```diff
@@ -33,12 +32,28 @@ def __init__(self, *args, **kwargs):
...
                 index = obj["index"]
             return pd.Series(obj["data"], index=index)
+
+        # Check for DataFrame
+        if "__type__" in obj and obj["__type__"] == "pd.DataFrame":
+            if obj.get("is_multiindex", False):
+                index = pd.MultiIndex.from_tuples(
+                    obj["index"], names=obj.get("index_names")
+                )
+                return pd.DataFrame(obj["data"], columns=obj["columns"], index=index)
```
christina-moore, line 51: Entirely possible this is just `pandas` notation I don't understand 😅 but this conditional looks like if `obj` is not multiindex, then it is read as MultiIndex?

ssda903/multinomial.py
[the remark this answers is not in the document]
christina-moore: It's a great train of thought to follow through on - how could we end up in this situation and what do we want to do about it? Maybe we want to surface an error to the end user that they can then report to us?

dm_regional_app/templatetags/table_tags.py
christina-moore: Ah ok, and why is the html built here rather than using a template and passing data to fill in the template?

dm_regional_app/forms.py
amynickolls: Happy to chat through! Essentially:

This form handles entry rates, transition rates, and exit rates.

Exit rates are a subsection of transition rates as these are considered in the model as a transition to 'not in care'. These rates are always applied to a daily population (eg if fostering to residential is 0.1, 10% of the fostering populated will move to residential each day). The dataframes for transition rates will always be multiindex as they always have a 'from' bin and a 'to' bin.

Entry rates are basically just numbers. e.g. if the fostering entry rate is 0.1 then 0.1 of a child will enter each day. These dataframes will be single index as they only have the 'to' bin information.

dm_regional_app/utils.py
```diff
@@ -33,12 +32,28 @@ def __init__(self, *args, **kwargs):
 
     def parse_object(self, obj):
         obj = self.parse_dates(obj)
+        # Check for Series
         if "__type__" in obj and obj["__type__"] == "pd.Series":
             if obj.get("is_multiindex", False):
-                index = pd.MultiIndex.from_tuples(obj["index"])
+                index = pd.MultiIndex.from_tuples(
+                    obj["index"], names=obj.get("index_names")
+                )
```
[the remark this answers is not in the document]
amynickolls: It's now only adjusted_proportions and adjusted_costs that are series now - none will actually be multiindex now that the rates will be dataframes. We can chat through this in more detail!

dm_regional_app/utils.py
amynickolls: I had thought this but as it was working as intended left it. On further investigation it was working because (I think) the check was just seeing if it could get "is_multiindex" as when I changed the logic to `if obj.get("is_multiindex") == True:` it stopped working. No idea why it wasn't breaking on single index series/dfs.

On investigating this I'm noticing some weird behaviour - whenever I'm saving to the scenario, anything I've already saved is rerunning through the encoder. Is this expected behaviour?

e.g I'm saving adjusted rates but have previously saved entry rates - this will encode both items? Might be worth discussing!

christina-moore: Hmm yeah sounds like it's worth discussing this one too!

dm_regional_app/templatetags/table_tags.py
christina-moore: As discussed on our call, happy for this to be a refactor in another ticket that we can tackle later 🙂

amynickolls: I think both of us might have made an issue for this in different words

dm_regional_app/utils.py
```diff
@@ -33,12 +32,29 @@ def __init__(self, *args, **kwargs):
 
     def parse_object(self, obj):
         obj = self.parse_dates(obj)
+        # Check for Series
         if "__type__" in obj and obj["__type__"] == "pd.Series":
-            if obj.get("is_multiindex", False):
-                index = pd.MultiIndex.from_tuples(obj["index"])
+            if obj.get("is_multiindex") == True:
```
christina-moore, line 37: You can do 
```suggestion
            if obj.get("is_multiindex"):
```
but up to you on what you think is more readable!

dm_regional_app/utils.py
```diff
@@ -33,12 +32,29 @@ def __init__(self, *args, **kwargs):
...
+                index = pd.MultiIndex.from_tuples(
+                    obj["index"], names=obj.get("index_names")
+                )
             else:
                 index = obj["index"]
             return pd.Series(obj["data"], index=index)
+
+        # Check for DataFrame
+        if "__type__" in obj and obj["__type__"] == "pd.DataFrame":
+            print("saved", obj.get("is_multiindex"))
```
christina-moore, line 47: Should this be removed?

dm_regional_app/utils.py
```diff
@@ -65,6 +81,41 @@ def default(self, obj):
...
+        if isinstance(obj, pd.DataFrame):
+            # Handle MultiIndex DataFrame
+            if isinstance(obj.index, pd.MultiIndex):
+                index = [
+                    list(tup) for tup in obj.index
+                ]  # Convert MultiIndex to list of lists
+                index_names = obj.index.names  # Get index names
+            else:
+                index = obj.index.tolist()
+                index_names = obj.index.names  # Get index names
```
christina-moore, line 94: There's more of a pythonic question here around separation of concerns.

As python is not a typed language (but we do introduce some via the `typing` library) when there are type checks within functions (both here and above) it makes me think about if we are trying to do "too much" with one class. I'm curious for your thoughts on having two encoders and decoders - one that handles series and one that handles dataframes. In my mind that would help with clarity i.e. `DataFrameAwareJSONEncoder` vs `SeriesAwareJSONEncoder`. And could help with tests as well - more specific tests to target individual functionality.

christina-moore: @cyramic I would also be curious for your thoughts here. 
@amynickolls I don't see this as a blocker to the code merge - just an interesting question that could shape some of our design decisions going forward!

dm_regional_app/utils.py
```diff
@@ -65,6 +81,41 @@ def default(self, obj):
...
+            # Transform NaN to None only when necessary
+            records = []
+            for record in obj.to_dict(orient="records"):
+                transformed_record = {
+                    key: (value if pd.notnull(value) else None)
+                    for key, value in record.items()
+                }
+                records.append(transformed_record)
+            columns = obj.columns.tolist()  # Get column names
+            print("is multi index", isinstance(obj.index, pd.MultiIndex))
```
christina-moore, line 105: Should this be removed? And the print below?

dm_regional_app/utils.py
```diff
@@ -86,3 +137,27 @@ def number_format(value):
...
+    # start new data dataframe
+    new_data = saved_data.copy()
+
+    # for rows where both form_data and saved_data have the same index
+    for idx in form_data.index.intersection(saved_data.index):
+        # replace saved data with new data
+        new_data.loc[idx] = form_data.loc[idx]
+
+    # Select unmatched rows from form_data that are not in saved_data
+    unmatched_form_data = form_data[~form_data.index.isin(saved_data.index)]
```
christina-moore, line 158: Just for my education - what is the `~` doing here?

dm_regional_app/views.py
```diff
@@ -392,11 +397,11 @@ def placement_proportions(request):
                     new_numbers = data.combine_first(proportion_adjustments)
 
                     session_scenario.adjusted_proportions = new_numbers
-                    session_scenario.save()
+                    session_scenario.save(update_fields=["adjusted_proportions"])
 
                 else:
                     session_scenario.adjusted_proportions = data
-                    session_scenario.save()
+                    session_scenario.save(update_fields=["adjusted_proportions"])
```
christina-moore, line 404: Did these end up reducing the doubled encoding?

dm_regional_app/utils.py
christina-moore: Ah of course, Thursday morning brain had me thinking it was a dataframe-specific thing 😅

dm_regional_app/utils.py
amynickolls: I wouldn't mind splitting them - they do handle dates as well which I _think_ would need to be a secondary function of each en/decoder anyway (although I can test this - it's possible that this was mainly an issue in the dictionaries, but assume it would be the same for dates in columns).

cyramic: I agree that separating them would make them more maintainable and more easily testable. I'm also thinking that since they are similar structures there may be an argument for maybe a design pattern to do this such that all encoders defined of this type follow a standard pattern. Maybe something like a [Builder](https://refactoring.guru/design-patterns/builder/python/example#example-0) or maybe one of the Factory patterns?  That might be overdoing it though, so happy to get push back as I don't want to over-engineer this. What do you think @christina-moore?

christina-moore: I'm happy for this to be pulled into another PR/issue for clarity

ssda903/multinomial.py
MichaelHanksSF, line 1: Happy with this - thanks

dm_regional_app/views.py
MichaelHanksSF, line 1: Happy with the change to a function - thanks!

MichaelHanksSF (approved):
Happy with this!

PR #97, opened 2024-10-04
39 update historic data filter

amynickolls:
Updates to historic data filter:
- Ethnicity and Sex filters added
- Age and Placement Type filters removed
- UASC filtering logic fixed

To complete this:
- Datacontainer work to add Ethnicity column
- UASC logic in datacontainer updated

Additional updates needed due to historic filter change:
- router_handler filter defaults
- custom_tags update (to display filters as text on costs page)

### DIFF
--- a/dm_regional_app/templatetags/custom_tags.py
+++ b/dm_regional_app/templatetags/custom_tags.py
@@ -48,7 +48,12 @@ def filter_text(dict):
     # Only include 'uasc' if its value is not 'all' and not empty
     if dict.get("uasc") and dict["uasc"] != "all":
         html += "UASC:<ul>"
-        html += f"<li>{dict['uasc']}</li></ul>"
+        # Check the value of dict["uasc"]
+        if dict["uasc"] == "True":
+            html += "<li>UASC only</li>"
+        else:
+            html += "<li>UASC Excluded</li>"
+        html += "</ul>"
 
     if (
         html
--- a/dm_regional_app/templatetags/custom_tags.py
+++ b/dm_regional_app/templatetags/custom_tags.py
@@ -28,11 +28,21 @@ def filter_text(dict):
             html += f"<li>{la}</li>"
         html += "</ul>"
 
-    # Check for 'age_bins' and ensure it has non-empty values
-    if dict.get("age_bins"):
-        html += "Age bins:<ul>"
-        for age_bin in dict["age_bins"]:
+    # Check for 'ethnicity' and ensure it has non-empty values
+    if dict.get("ethnicity"):
+        html += "Ethnicity:<ul>"
+        for age_bin in dict["ethnicity"]:
             html += f"<li>{age_bin}</li>"
+        html += "</ul>"
+
+    # Only include 'sex' if its value is not 'all' and not empty
+    if dict.get("sex") and dict["sex"] != "all":
+        html += "Sex:<ul>"
+        # Check the value of dict["sex"]
+        if dict["sex"] == "1":
+            html += "<li>Male</li>"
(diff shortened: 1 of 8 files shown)


dm_regional_app/templatetags/custom_tags.py
```diff
@@ -28,13 +28,23 @@ def filter_text(dict):
...
+    # Only include 'sex' if its value is not 'all' and not empty
+    if dict.get("sex") and dict["sex"] != "all":
+        html += "Sex:<ul>"
+        # Check the value of dict["sex"]
+        if dict["sex"] == "1":
+            html += "<li>Male</li>"
+        else:
+            html += "<li>Female</li>"
+        html += "</ul>"
+
```
MichaelHanksSF, line 47: Is there any reason why lines 42-26 are not done like with UASC below?

dm_regional_app/utils.py
```diff
@@ -15,18 +13,19 @@ def apply_filters(data: pd.DataFrame, filters: dict):
         loc = data.LA.astype(str).isin(filters["la"])
         data = data.loc[loc]
 
-    if filters["placement_types"] != []:
-        loc = data.placement_type.astype(str).isin(filters["placement_types"])
+    if filters["ethnicity"] != []:
+        loc = data.ethnicity.astype(str).isin(filters["la"])
```
MichaelHanksSF, line 17: This is incorrect - should read filters["ethnicity"]

dm_regional_app/views.py
```diff
@@ -67,8 +67,8 @@ def router_handler(request):
 
     historic_filters = {
         "la": [],
-        "placement_types": [],
-        "age_bins": [],
+        "ethnicity": [],
+        "sex": [],
```
MichaelHanksSF, line 71: I'm not 100%, but shouldn't this be "sex": "all"?

dm_regional_app/views.py
```diff
@@ -935,8 +935,7 @@ def adjusted(request):
                 historic_form = HistoricDataFilter(
                     request.POST,
                     la=datacontainer.unique_las,
-                    placement_types=datacontainer.unique_placement_types,
-                    age_bins=datacontainer.unique_age_bins,
+                    ethnicity=datacontainer.unique_ethnicity,
```
MichaelHanksSF, line 938: Just checking my understanding, but are sex and uasc not here because they have a fixed set of choices?

MichaelHanksSF (changes requested):
Thanks for this!

Couple of changes I think are required.

dm_regional_app/templatetags/custom_tags.py
amynickolls: UASC flag in the data is True and False which is understandable to users, however, might be worth updating the UASC code so it reads the same as the filter (Include/Exclude UASC)?

dm_regional_app/views.py
amynickolls: Yes - placement types and age bins are read from the data and input to the form which allows for users to change these in the datacontainer without needing to update the Django app. We could treat sex in the same way if we wanted, adding an extra column in data container for text description? This might help future-proof it if DFE were to update sex to include gender identity/markers for trans/gender non-conforming children in future.

amynickolls: But then it's likely that if they were to update to beyond binary we would want to change the filter to multiselect anyway...

MichaelHanksSF: Yep - makes sense. I think leave as is.

MichaelHanksSF (approved):

PR #98, opened 2024-10-07
Fix error - empty form causing empty dataframe JSON issue

amynickolls:
This fixes the error created by saving an empty form and then navigating away.

The error was caused by the JSON decoder being unable to handle an empty dataframe. We could introduce code to enable the encoder to handle this, however this would likely lead to the empty dataframe causing further errors further down the chain. We should think about whether this is worth implementing despite the following fix.

Have implemented form to only save in initial instance (a user has not saved rates/numbers/proportions before) if the dataframe from the form is not empty.

### DIFF
--- a/dm_regional_app/utils.py
+++ b/dm_regional_app/utils.py
@@ -1,10 +1,8 @@
 import ast
 import json
-import re
 from datetime import date, datetime
 
 import pandas as pd
-import plotly.express as px
 import plotly.graph_objects as go
 
 from ssda903.config import PlacementCategories
@@ -214,3 +212,12 @@ def add_ci_traces(df, fig):
         )
 
     return fig
+
+
+def save_data_if_not_empty(session_scenario, data, attribute_name):
+    """
+    Checks if series or dataframe is not empty, and saves to attribute of model if not
+    """
+    if isinstance(data, (pd.DataFrame, pd.Series)) and not data.empty:
+        setattr(session_scenario, attribute_name, data)
+        session_scenario.save()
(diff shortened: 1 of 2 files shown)


tab1tha (commented):
Apart from the session _scenario attribute changes, the four lines look like they're the same (I could be wrong). Could this be put in a function that receives the attribute name as an argument, so that doesn't repeat multiple times? 
This might bring it closer to the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) development principle
e.g 

> def safe_save(scenario_attribute):

tab1tha:
> Apart from the session _scenario attribute changes, the four lines look like they're the same (I could be wrong). Could this be put in a function that receives the attribute name as an argument, so that doesn't repeat multiple times? This might bring it closer to the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) development principle e.g
> 
> > def safe_save(scenario_attribute):

On second thought, since an attribute cannot be simply string-joined to its object, there is a possibility that wrapping this in a function might become tricky and make it less readable. 
This is sample code that I got from a chatbot:
```

def save_data_if_exists(session_scenario, data, attribute_name):
    if isinstance(data, (pd.DataFrame, pd.Series)) and not data.empty:
        setattr(session_scenario, attribute_name, data)
        session_scenario.save()
# Example usage:
# save_data_if_exists(session_scenario, data, 'adjusted_costs')

```
It might work but if this makes the code less intuitive then the current implementation would be better. So, I'll approve this and let you decide.

tab1tha (approved):

amynickolls:
@tab1tha your suggestion worked so I've gone with that! Think there is a lot in these views that could better follow DRY principles...

PR #108, opened 2024-10-22
rate change pages return user to previous page

### LINKED ISSUE #92: Edit rate adjustment logic to return to costs/adjusted based on initial page
All rate adjustment pages (transition, entry and exit) are accessible via both Adjusted and Costs pages. Add logic so that the return page is passed through the URL, and logic for displaying return to costs or adjusted page (at present all returns to adjusted).

WillLP-code:
closes #92 

When accessing entry_rates, transition_rates, and exit_rates, the back/return to x page/clear all rate adjustments buttons now returns the user to the page they accessed them from (adjusted or costs).

The obvious choice would be to set prev_page with request.META.HTTP_REFERER but when you save your changes to a transition rate table, the previous page is the page containing the table (e.g. if we save the entry_rates page, request.META.HTTP_REFERER gets entry_rates not adjusted/costs).

I'm not certain saving it in the session as "prev_page" is a good key name as it's not always the previous page, but I'd like guidance on whether I'm tacking this the right way before I figure out what to do here. I could delete the session key once the prev_page variable is set on a given rate change page but I'd need to work out where to do it.


christina-moore:
Using the session makes sense with the code as it stands (in that it looks like it is intentionally done such that saving the form does not automatically return the user to a previous page). From a naming perspective, perhaps something more verbose could be useful e.g. `entry_rates_origin_page`. And then in the html I think we'll need something cleaner in the button (as I anticipate the current `{{prev_page}}` would output a button name that looks more like a URL segment)

WillLP-code:
@christina-moore Thanks for the feedback. 

I've changed the `prev_page` variable to `rate_change_origin_page` based on your suggestion which I think is a better explanation of of the variable.

I've also changed the logic on the button that returns you to the previous page from the rate adjustment pages to have a better button text. I have two points/questions(?) on this:
1. Is it okay to put the logic in the html template, or would it be better to have it in the views and pass the button text as a variable to the template in the session as I did the `rate_change_origin_page` string?
2. I've used lower case 'adjust forecast' and 'projecting spend' text on the buttons, in keeping with other buttons on the site, but I've also added the word page to the end which isn't on buttons elsewhere in the site because it felt more syntactically correct, but I'm no grammarian, and I don't know if there is a best practice for button naming, I'd love to hear suggestions?

.env.example
christina-moore, line 1: Should this file have been deleted?

dm_regional_app/views.py
```diff
@@ -119,6 +119,9 @@ def costs(request):
         pk = request.session["session_scenario_id"]
         session_scenario = get_object_or_404(SessionScenario, pk=pk)
 
+        # Used to return user to this page when accessing rate change pages
+        request.session["rate_change_origin_page"] = "costs"
```
christina-moore, line 123: Could be worth using [this function](https://docs.djangoproject.com/en/5.1/ref/urlresolvers/#reverse) here as it would reduce hardcoding

dm_regional_app/templates/dm_regional_app/views/exit_rates.html
```diff
@@ -43,9 +43,15 @@ <h1>Adjust exit rates</h1>
 </div>
 
 <div class="d-flex bd-highlight mb-3">
-    <div class="p-2 bd-highlight"><a class="btn btn-warning" href="{% url 'clear_rates' %}?next_url_name=adjusted">Clear all rate adjustments</a></div>
-    <div class="ms-auto p-2 bd-highlight"><a class="btn btn-primary" href="{% url 'router_handler' %}?next_url_name=adjusted">Return to adjusted forecast</a></div>
-  </div>
+    <div class="p-2 bd-highlight"><a class="btn btn-warning" href="{% url 'clear_rates' %}?next_url_name={{rate_change_origin_page}}">Clear all rate adjustments</a></div>
+    <div class="ms-auto p-2 bd-highlight"><a class="btn btn-primary" href="{% url 'router_handler' %}?next_url_name={{rate_change_origin_page}}">
+        {% if rate_change_origin_page == 'costs' %}
+            Return to projecting spend page
```
christina-moore, line 49: My personal preference is omitting "page" as that feels implied by the existence of the pages from whence (I've never used that in a sentence) the user came. 
I think overall lower case is fine as you're right, it's a site-wide custom. @MichaelHanksSF any opinions on the use of the word "page" here?

dm_regional_app/templates/dm_regional_app/views/exit_rates.html
```diff
@@ -43,9 +43,15 @@ <h1>Adjust exit rates</h1>
...
+        {% elif rate_change_origin_page == 'adjusted' %}
```
christina-moore, line 50: As for the conditional, it's dealer's choice really whether it lives in the template or the view. As this is quite a small conditional, I think keeping it in the template is fine.

dm_regional_app/templates/dm_regional_app/views/transition_rates.html
```diff
@@ -39,9 +39,15 @@ <h1>Adjust transition rates</h1>
 </div>
 
 <div class="d-flex bd-highlight mb-3">
-    <div class="p-2 bd-highlight"><a class="btn btn-warning" href="{% url 'clear_rates' %}?next_url_name=adjusted">Clear all rate adjustments</a></div>
-    <div class="ms-auto p-2 bd-highlight"><a class="btn btn-primary" href="{% url 'router_handler' %}?next_url_name=adjusted">Return to adjusted forecast</a></div>
-    </div>
+    <div class="p-2 bd-highlight"><a class="btn btn-warning" href="{% url 'clear_rates' %}?next_url_name={{rate_change_origin_page}}">Clear all rate adjustments</a></div>
```
christina-moore, line 42: It looks like `entry_rates`, `exit_rates`, and `transition_rates` are quite similar templates - I wonder if there is some DRYing that is worth doing here 🤔 Perhaps one for a new ticket/PR!

dm_regional_app/templates/dm_regional_app/views/exit_rates.html
```diff
@@ -43,9 +43,15 @@ <h1>Adjust exit rates</h1>
 </div>
 
 <div class="d-flex bd-highlight mb-3">
-    <div class="p-2 bd-highlight"><a class="btn btn-warning" href="{% url 'clear_rates' %}?next_url_name=adjusted">Clear all rate adjustments</a></div>
-    <div class="ms-auto p-2 bd-highlight"><a class="btn btn-primary" href="{% url 'router_handler' %}?next_url_name=adjusted">Return to adjusted forecast</a></div>
-  </div>
+    <div class="p-2 bd-highlight"><a class="btn btn-warning" href="{% url 'clear_rates' %}?next_url_name={{rate_change_origin_page}}">Clear all rate adjustments</a></div>
```
christina-moore, line 46: Just wanted to double check that the use of `reverse`, now that it has prepended the slash, won't break this `next_url_name`?

christina-moore (approved):

PR #109, opened 2024-10-23
adds model states to adjusted graph

### LINKED ISSUE #99: Adjusted graph shows model states
Update adjusted graph to show model states as with standard predict graph.

Colours should match predict graph

Potentially switch solid line to be adjusted and long dash to be original.

WillLP-code:
closes #99

Uses colours from predict graph, makes original forecast dashed as suggested in issue.

### DIFF
--- a/dm_regional_app/charts.py
+++ b/dm_regional_app/charts.py
@@ -228,20 +228,14 @@ def prediction_chart(
     forecast_care_by_type_dfs["type"] = "Base forecast"
     forecast_care_by_type_dfs["dash"] = None
 
-    # Add dictionary to list to be added to graph
-    traces_list = [historic_care_by_type_dfs, forecast_care_by_type_dfs]
-
     # Add forecast and historical traces
-    fig = add_traces(fig, traces_list)
+    fig = add_traces(fig, [historic_care_by_type_dfs, forecast_care_by_type_dfs])
 
     # Append graph info to ci data dictionaries
     df_ci["type"] = "Base forecast"
 
-    # Add dictionary to list to be added to graph
-    ci_traces_list = [df_ci]
-
     # Display confidence interval as filled shape
-    fig = add_ci_traces(fig, ci_traces_list)
+    fig = add_ci_traces(fig, [df_ci])
 
     # add shaded reference period
     fig.add_shape(
@@ -287,11 +281,8 @@ def historic_chart(data: PopulationStats) -> str:
     historic_care_by_type_dfs["type"] = "Historic"
     historic_care_by_type_dfs["dash"] = "dot"
 
-    # Add dictionary to list to be added to graph
-    traces_list = [historic_care_by_type_dfs]
-
     # Add historical traces
-    fig = add_traces(fig, traces_list)
+    fig = add_traces(fig, [historic_care_by_type_dfs])
 
     fig.update_layout(title="Historic child population over time")
     fig.update_yaxes(rangemode="tozero")
@@ -473,25 +464,22 @@ def compare_forecast(
     adjusted_care_by_type_dfs["type"] = "Adjusted forecast"
     adjusted_care_by_type_dfs["dash"] = "dash"
 
-    # Add dictionary to list to be added to graph
-    traces_list = [
-        historic_care_by_type_dfs,
-        forecast_care_by_type_dfs,
-        adjusted_care_by_type_dfs,
-    ]
-
     # Add historical and forecast data to figure
-    fig = add_traces(fig, traces_list)
+    fig = add_traces(
+        fig,
+        [
+            historic_care_by_type_dfs,
+            forecast_care_by_type_dfs,
+            adjusted_care_by_type_dfs,
+        ],
+    )
 
     # Append graph info to ci data dictionaries
     df_ci["type"] = "Base forecast"
     df_af_ci["type"] = "Adjusted forecast"
 
-    # Add dictionary to list to be added to graph
-    ci_traces_list = [df_ci, df_af_ci]
-
     # Display confidence interval as filled shape
-    fig = add_ci_traces(fig, ci_traces_list)
+    fig = add_ci_traces(fig, [df_ci, df_af_ci])
 
     # add shaded reference period
     fig.add_shape(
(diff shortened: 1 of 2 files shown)


MichaelHanksSF:
@WillLP-code I have made some amends to this, based on a couple of things:
- I added a lot of annotation to the functions that were relevant here, including some pre-existing ones, giving them docstrings, descriptions of inputs and outputs
- I removed add_original_traces as I thought this was an unnecessary repetition and just made the arguments to add_traces optional
- I applied the same kind of if base:, if adjusted: logic to add_ci_traces mostly to accomplish the ability to add legendgroups to the traces. This is because the legend was very cluttered and I thought it would be helpful for people to be able to hide confidence intervals at the same time they hid the main line.

Apologies for making these changes, rather than requesting them - to some extent I wasn't sure what was wanted until I did it (like the confidence intervals) so I guess I got carried away!

MichaelHanksSF:
@christina-moore Could you review the changes I've made to Will's PR please. I'd ask Amy, but she's going on holiday after tomorrow.

dm_regional_app/charts.py
```diff
@@ -189,23 +187,28 @@ def summary_tables(df):
...
+    """
     # Pop start and end dates to visualise reference period
     reference_start_date = kwargs.pop("reference_start_date")
     reference_end_date = kwargs.pop("reference_end_date")
 
     # Dataframe containing total children in prediction
     df = prediction.population.unstack().reset_index()
 
-    df.columns = ["from", "date", "forecast"]
+    df.columns = ["bin", "date", "forecast"]
```
christina-moore, line 203: Just curious about the move from "from" to "bin" for my own understanding?

dm_regional_app/utils.py
```diff
@@ -174,25 +177,35 @@ def care_type_organiser(df, data_type, input_col):
...
 
-def add_traces(dfs_forecast, dfs_historic, fig):
+def add_traces(
+    fig: go.Figure,
+    historic_by_type: Optional[dict] = None,
+    forecast_by_type: Optional[dict] = None,
+    adjusted_by_type: Optional[dict] = None,
+) -> go.Figure:
+    """
+    Adds populations (forecast and historic) to graph, split by five categories e.g. placement categories
```
christina-moore, line 207: Should this include `adjusted`?

dm_regional_app/utils.py
```diff
@@ -243,34 +278,61 @@ def add_ci_traces(df, fig):
...
+            # Add lower bounds for confidence intervals.
+            fig.add_trace(
+                go.Scatter(
+                    x=adjusted_ci_dict[care_type]["date"],
+                    y=adjusted_ci_dict[care_type]["lower"],
+                    line_color="rgba(255,255,255,0)",
+                    legendgroup=f"Adjusted {care_type}",
+                    showlegend=False,
+                )
+            )
```
christina-moore, line 318: Does it make sense to pull the loop out to the top level and use the conditionals inside so we iterate once over the loop? The answer could be no if we think it makes more sense for readability to keep them separate.
Also am curious what the different between this function and `add_traces` are? They seem functionally very similar, but perhaps it makes sense to keep them separate for clarity?

christina-moore (commented):
Love more docstrings and typing ❤️

dm_regional_app/charts.py
MichaelHanksSF: In the context of the model, I think it's confusing to use "from" here. At this point, we're not talking about a transition, where the child is moving "from" one bin "to" another, where these terms would be appropriate (and are used elsewhere). Instead, we're just reporting on the population size of the bin. Therefore, to avoid confusion, I'd rather call these data fields "bin" rather than "from".

dm_regional_app/utils.py
MichaelHanksSF: Yes, it does. I agree with the similarity of add_traces and add_ci_traces and dissatisfied by the repetition within these as well. I think there's a lot more abstraction we could achieve with this set of functions. The only real difference between the cis and the other traces is the fill for the lower CI, but I would have thought we could just say none for the others. Then they can be in the same function.

Even better would be if the variation in graph inputs (x, y, legendgroup etc.) can be fed to the function in the input info. Then we could simply pass a single function a list of one or more sets of info and it could loop over these and produce the lines required.

Better still would be if we can abstract the enum that we are using. The difficulty here is the removal of "not in care" and the addition of "total". But "not in care" maybe could be removed before we pass to the function.

I'll have a look and see if we can make this work!

christina-moore: There is always a balance between DRY and readability/clarity (DRY-est is not always the most easily understood solution) so happy to have that wherever you all think best in this instance.

MichaelHanksSF:
@christina-moore ready for you review once more

MichaelHanksSF:
I abstracted the dictionaries going into each of the two trace functions, but kept those functions separate, mainly because the ci dictionaries have two columns of data (upper and lower) that need to be treated differently, so this doesn't combine as well with the basic population dictionaries (the historic data and forecasts).

I left the enum alone, as there's too much specificity in what we're doing with it, mostly due to removing "Not in care" from the population and the enum.

dm_regional_app/charts.py
```diff
@@ -218,11 +221,27 @@ def prediction_chart(historic_data: PopulationStats, prediction: Prediction, **k
...
+
+    forecast_care_by_type_dfs["type"] = "Base forecast"
+    forecast_care_by_type_dfs["dash"] = None
+
+    # Add dictionary to list to be added to graph
+    traces_list = [historic_care_by_type_dfs, forecast_care_by_type_dfs]
+
     # Add forecast and historical traces
-    fig = add_traces(forecast_care_by_type_dfs, historic_care_by_type_dfs, fig)
+    fig = add_traces(fig, traces_list)
```
christina-moore, line 235: This is up to personal preference. I usually prefer not to define variables if they aren't being re-used. So here I would opt for:
```suggestion
    fig = add_traces(fig, [historic_care_by_type_dfs, forecast_care_by_type_dfs])
```
The idea being that creating a variable could imply the need for re-use which adds to developer mental load when reading/modifying the code. [This readme](https://github.com/zakirullin/cognitive-load?tab=readme-ov-file#cognitive-load) has some interesting other practices around reducing this 🙂

MichaelHanksSF: I can get behind that idea - thanks!

christina-moore (approved):

PR #132, opened 2024-11-05
feat: use middleware to replace router handler

christina-moore:
This update will remove the `router_handler` function and instead use a custom middleware. Middleware is applied to every request/response so now all views will have a `session_scenario_id` stored in the session. 

`views.py` changes look extensive, but for each function all I did was remove the conditional that checks for the id in the session. Probably easiest to see those changes within this commit itself: https://github.com/SocialFinanceDigitalLabs/cs-demand-model-regional/commit/5f7c6c0af2ea32d5426858fc1a8bac59c0b35d23

Note: If there are no data uploads, the middleware will redirect the user to the home screen with a message:
image

### DIFF
--- a/dm_regional_app/middleware/scenario_middleware.py
+++ b/dm_regional_app/middleware/scenario_middleware.py
@@ -1,7 +1,17 @@
+import re
+
+from django.contrib import messages
+from django.shortcuts import redirect
+
 from dm_regional_app.models import DataSource, SessionScenario
 
 
 class SessionScenarioMiddleware:
+    PATHS_TO_IGNORE = re.compile(
+        r"(/admin.*)|(/accounts.*)|(/upload_data.*)|(/$)",
+        re.IGNORECASE,
+    )
+
     def __init__(self, get_response):
         self.get_response = get_response
 
@@ -9,7 +19,11 @@ class SessionScenarioMiddleware:
         # Initialise a SessionScenario
         session_scenario_id = request.session.get("session_scenario_id", None)
         current_user = request.user
-        if current_user.is_authenticated and DataSource.objects.exists():
+
+        if not current_user.is_authenticated:
+            return self.get_response(request)
+
+        if DataSource.objects.exists():
             historic_filters = {
                 "la": [],
                 "ethnicity": [],
@@ -49,7 +63,15 @@ class SessionScenarioMiddleware:
             )
 
             request.session["session_scenario_id"] = session_scenario.pk
+        else:
+            if self.PATHS_TO_IGNORE.match(request.path_info):
+                return self.get_response(request)
 
-        response = self.get_response(request)
+            messages.error(
+                request,
+                "There is no data available. If you are an admin, please upload data via Data Source Upload. Otherwise, "
+                "please contact your local admin.",
+            )
+            return redirect("home")
 
-        return response
+        return self.get_response(request)
--- a/dm_regional_app/middleware/scenario_middleware.py
+++ b/dm_regional_app/middleware/scenario_middleware.py
@@ -1,0 +1,55 @@
+from dm_regional_app.models import DataSource, SessionScenario
+
+
+class SessionScenarioMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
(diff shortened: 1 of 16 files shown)


dm_regional_app/middleware/scenario_middleware.py
```diff
@@ -0,0 +1,55 @@
...
+            historic_stock = {
+                "population": {},
+                "base_rates": [],
+            }
+            inflation_parameters = {
+                "inflation": False,
+                "inflation_rate": 0.1,
+            }
+
+            # default_values should define the model default parameters, like reference_date and the stock data and so on. Decide what should be default with Michael
```
amynickolls, line 36: I think the second half of this comment is no longer relevant

dm_regional_app/tests/test_data_upload.py
```diff
@@ -57,7 +58,13 @@ def test_data_upload_success(self, prediction):
             "header": SimpleUploadedFile("header.csv", b"header"),
             "uasc": SimpleUploadedFile("uasc.csv", b"uasc"),
         }
-        prediction.return_value = True, None
+
+        # Mock the datacontainer created from the uploaded files
+        datacontainer = MagicMock()
+        type(datacontainer).start_date = datetime(2024, 1, 1)
+        type(datacontainer).end_date = datetime(2024, 12, 1)
```
tab1tha, line 65: I'm not well-versed with Mock syntax so I have a question about this: Is datacontainer here a mock of DataSource?

tab1tha (approved):

dm_regional_app/middleware/scenario_middleware.py
```diff
@@ -0,0 +1,55 @@
...
+        current_user = request.user
+        if current_user.is_authenticated and DataSource.objects.exists():
+            historic_filters = {
+                "la": [],
+                "ethnicity": [],
+                "sex": "all",
+                "uasc": "all",
+            }
+
+            data = DataSource.objects.order_by("-uploaded").first()
```
tab1tha, line 20: Does this line get the most recently created DataSource object? If so, I just learnt from @amynickolls's [pull request](https://github.com/SocialFinanceDigitalLabs/cs-demand-model-regional/pull/133#discussion_r1838278878) that there is a [`.latest` method](https://docs.djangoproject.com/en/dev/ref/models/querysets/#latest) that seems to do this too.
> data = DataSource.objects.latest("uploaded")

christina-moore:
> Should load_saved_scenario view be updated to use middleware to store session ID, I believe it's still using the old method?

Hmm so the middleware can't be redirected to like other views. On each request made by a user, the middleware will "intercept" the request and then add it's logic. I think it's fine to still have `load_saved_scenario` set the scenario ID into the session directly - as long as the middleware isn't overwriting what `load_saved_scenario` is doing! So that's one to test for sure

dm_regional_app/middleware/scenario_middleware.py
christina-moore: Have updated: https://github.com/SocialFinanceDigitalLabs/cs-demand-model-regional/pull/132/commits/843e7091810262ccc09682d72bc4c0f594b985c7

dm_regional_app/middleware/scenario_middleware.py
christina-moore: Good shout - this one is new to me. https://github.com/SocialFinanceDigitalLabs/cs-demand-model-regional/pull/132/commits/843e7091810262ccc09682d72bc4c0f594b985c7

dm_regional_app/tests/test_data_upload.py
christina-moore: Ah so `datacontainer` is a mock of the object that is returned by `validate_with_prediction`. Basically what it's doing is creating a fake datacontainer object with a start_date and end_date and then instead of running `validate_with_prediction` at all, it's mocking that method and returning from it the fake object I created. The comment seemed a little misleading so I've updated it as well: https://github.com/SocialFinanceDigitalLabs/cs-demand-model-regional/pull/132/commits/575e86491f2969d124f0a6b909aefb15e73d6f07

PR #133, opened 2024-11-12
70 implement explainer page after landing page

amynickolls:
Homepage for authenticated users added.
image


dm_regional_app/templates/dm_regional_app/views/home.html
```diff
@@ -2,6 +2,8 @@
 {% block content %}
     {% if not user.is_authenticated %}
         {% include "dm_regional_app/includes/guest_home.html" %}
+    {% elif user.is_authenticated %}
+        {% include "dm_regional_app/includes/authenticated_home.html" %}
```
christina-moore, line 6: Could use just an `else` here as `user.is_authenticated` should be a boolean. Unless there is a None case we need to handle 🤔

tab1tha:
side note with regard to the commit history: There are two adjacent commits which correct a typo. One other way to do this is that when you notice a change that should have been done in your previous commit, you can add it to that previous commit by doing `git commit --amend` instead. 
A window will pop up with the merge message. It can be closed by typing `:wq` or `esc then :wq` and it edits the local previous commit.
Then do `git push origin head --force-with-lease` to edit the remote previous commit too.

amynickolls:
thanks! I need to get better at using git in the terminal - I've saved your instructions for future ref

PR #143, opened 2024-12-03
feat: use middleware to check for password update

christina-moore:


dm_regional_app/middleware/update_password_middleware.py
```diff
@@ -0,0 +1,34 @@
...
+
+class UpdatePasswordMiddleware:
+    """
+    This middleware will check if a user needs to update their password and redirect them
+    accordingly. If the user's account is not using SSO and this is their first time logging
+    in to the application, then they must change their password.
+    """
+
+    PATHS_TO_IGNORE = re.compile(
+        r"(/admin.*)|(/accounts.*)",
```
cyramic, line 14: While i don''t have any specific concerns, I worry that this regex is too permissive. It would allow things such as:

* /accounts123
* /admin_1
* /admin/../../bad_thing/

While I think that django protects against these kinds of things, I wonder if it'd be worth making this more specific and less wildcardy (that's a word now)?

cyramic (commented):
Have tested locally, and the password function works as intended from my tests.

dm_regional_app/middleware/update_password_middleware.py
christina-moore: That's definitely a word now! 
The reason I'm not too worried about it is that this regex is defining endpoints/views that users can access without changing their password. If we don't exclude all the accounts views, then they will be stuck in an infinite loop potentially. All the admin views are excluded because only those with SSO will be made admins and they won't have passwords anyway (no one else will be able to access the admin endpoints). If an endpoint like `/admin/../../bad_thing/` did exist, I don't think catching it here would provide much help as this would just require a password change before navigating to that page 🤔 

That being said, I could see the case for it being `accounts/` including the trailing slash in case we want to name another app starting with `accounts...`

cyramic: In that case, I'm happy to approve as I don't see any paths that could accidentally be caught up in this.

cyramic (approved):

PR #145, opened 2024-12-17
Historic data and pop stats

MichaelHanksSF:
Few things going on here:
- [ ] removal of generic start_date and end_date variable naming in favour of either:
	- [ ] data_start/end_date
	- [ ] reference_start/end_date
	- [ ] prediction_start/end_date
EXCEPT where it's just done inside a function and never called from outside that function
- [ ] update PopulationStats to require a data_start_date and data_end_date as a property that it always gets from datacontainer.data_start_date and datacontainer.data_end_date; this is for use in stock and transitions to truncate the data to be contained within these dates; primary function to reduce the scope of the dates in the charts
- [ ] amended predictor.predict() to take PopulationStats as an input, rather than datacontainer.enriched_data, so that data_start_date and data_end_date do not need to be passed to predict() every time just so the first step can be stats = PopulationStats(data)
- [ ] removal of population_stats.to_excel; not clear what the function of this feature is
- [ ] removal of population_stats.ageing_out; this functionality has been superseded by explicitly considering ageing as transitions
- [ ] update logic in entry to care and exit from care box stats in historic view

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,6 +5,7 @@ description = ""
 authors = ["kws", "franciscobmacedo", "amynickolls", "MichaelHanksSF"]
 readme = "README.md"
 packages = []
+package-mode = false
 
 [tool.poetry.dependencies]
 python = "^3.10"


amynickolls (approved):
This looks good to me - my only comment is that there are some places (mostly within PopStats and then Predictor/Multinomial) where we are just using start_date or end_date and we could carry through the more descriptive start_date/end_date. However I think that the extent to which you have changed things can also be taken as a logical end point - most of these variables pull from a more descriptive version anyway, some of them I wouldn't change as the date becomes abstracted anyway (e.g. in a loop).

Approved as I don't think this is a barrier to merging but let me know if you want to go through these in more detail!

PR #146, opened 2024-12-18
remove age transition episodes from placement starts graph

MichaelHanksSF:
Simple one - small change to datacontainer to better identify episodes that are created to represent age transitions; then change in placement_starts_chart to remove these episodes from the graph


amynickolls (approved):
Seems sensible to me!

PR #148, opened 2025-01-08
Limit predict start date

MichaelHanksSF:
Small change to update PredictFilter validation on predict start date to not allow dates later than the datacontainer end date.


amynickolls (approved):
Do we want to make the max date the reference_end_date + 1 day?

PR #155, opened 2025-01-16
applies default of prediction start date to data end date

amynickolls:
Changes prediction start date default from None to data end date
Removes default in predict to reference end date


MichaelHanksSF (changes requested):
Apart from the predict variable inputs, I think this all works fine. Thanks!

MichaelHanksSF (approved):
Thank you!

PR #163, opened 2025-01-27
110 rate normalisation in multinomial

amynickolls:
Normalises rates when combining rates in multinomial, prioritising adjustments over unadjusted rates
Restricts users from adding more than 1 to a rate
Updates explanatory text to communicate changes

Rates summing to less than 1 are handled by `populate_same_state_transition`. This replaces the same state transition with 1-all other related rates, even if the same state transition already exists.

### DIFF
--- a/dm_regional_app/forms.py
+++ b/dm_regional_app/forms.py
@@ -410,6 +410,13 @@ class DynamicRateForm(forms.Form):
                     "You cannot multiply a rate if the original rate is 0",
                 )
 
+            # Validation logic: add value can't be more than 1
+            if add_value is not None and add_value > 1:
+                self.add_error(
+                    add_field_name,
+                    "You cannot add more than 1 to a rate",
+                )
+
         return cleaned_data
 
     def save(self):
--- a/dm_regional_app/forms.py
+++ b/dm_regional_app/forms.py
@@ -410,12 +410,13 @@ class DynamicRateForm(forms.Form):
                     "You cannot multiply a rate if the original rate is 0",
                 )
 
-            # Validation logic: add value can't be more than 1
-            if add_value is not None and add_value > 1:
-                self.add_error(
-                    add_field_name,
-                    "You cannot add more than 1 to a rate",
-                )
+            # Validation logic: add value can't be more than 1 for rates
+            if self.series.name != "daily_entry_probability":
+                if add_value is not None and add_value > 1:
+                    self.add_error(
+                        add_field_name,
+                        "You cannot add more than 1 to a rate",
+                    )
 
         return cleaned_data
 
--- a/dm_regional_app/templates/dm_regional_app/views/entry_rates.html
+++ b/dm_regional_app/templates/dm_regional_app/views/entry_rates.html
@@ -8,12 +8,16 @@
 <br>
     <h1>Adjust entry rates</h1>
     <br>
-    <p>You may want to edit the number of children per year entering care to give a more accurate
-        picture of future forecasting or simply to explore what changes you may expect to see. </p>
-    <p>Rates are applied to the daily population of children. For example, a rate of 0.5 to 10-16 Fostering 
-        would mean that every two days a child would enter this placement.
+    <p>
+        You may want to edit the number of children per year entering care to give a more accurate
+        picture of future forecasting or simply to explore what changes you may expect to see. 
     </p>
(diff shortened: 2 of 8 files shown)


ssda903/multinomial.py
```diff
@@ -31,13 +31,44 @@ class NextPrediction:
...
+            original[:] = 0
+        else:
+            # Scale down adjusted rates and distribute remaining capacity to original rates
+            adjusted = adjusted
+            remaining_capacity = 1 - adjusted.sum()
+            if not original.empty:
+                original = original * (remaining_capacity / original.sum())
+        # Combine adjusted and original back together
+        group = pd.concat([adjusted, original]).sort_index()
+    return group
```
MichaelHanksSF, line 59: I think this works well because my reading of this plus the existing populate_same_state_transitions is that the ageing rates won't be affected by other rate changes unless the other rate changes add up to more than 1-ageing_rate?

dm_regional_app/forms.py
```diff
@@ -410,6 +410,13 @@ def clean(self):
                     "You cannot multiply a rate if the original rate is 0",
                 )
 
+            # Validation logic: add value can't be more than 1
+            if add_value is not None and add_value > 1:
+                self.add_error(
+                    add_field_name,
+                    "You cannot add more than 1 to a rate",
+                )
+
```
MichaelHanksSF, line 419: I appreciate this adds complexity, but we do want entry rates to be able to be added to by more than 1. Is this possible?

MichaelHanksSF (changes requested):
Thanks for this - some good changes here.

Mostly all good, but we do want entry rates to allow >1 to be added. How much difficulty is this going to add?

ssda903/multinomial.py
MichaelHanksSF: I tested this and it's right - ageing is left alone by and large which is sensible given the choices we're giving people.

dm_regional_app/forms.py
amynickolls: Ah I see what you mean, I'll have a think!

dm_regional_app/templates/dm_regional_app/views/entry_rates.html
```diff
@@ -8,12 +8,16 @@
...
+        You may want to edit the number of children per year entering care to give a more accurate
+        picture of future forecasting or simply to explore what changes you may expect to see. 
     </p>
-    <p>Rate multiplication multiplies the initial base rate produced by the model parameters you selected.
+    <p>
+        Entry rates represent the number of children entering a placement each day. For example, a rate of 
+        0.5 in 10-16 Fostering would mean that every two days a child would enter this placement.
+    </p>
+    <p>
+        Adjustments are always applied to the initial base rate produced by the model parameters you selected, they do not replace them.
```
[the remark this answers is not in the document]
amynickolls: Maybe the distinction isn't needed - I wanted to communicate that if you add 0.1 to an exit rate and save, this then goes into a calculation of base rate + adjustment, rather than the base rate being replaced e.g. `base rate = base rate + 0.1`, and that saving again +0.1 won't mean that the rate now equals base rate + 0.2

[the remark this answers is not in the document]
amynickolls: I've deleted "they do not replace them" section for now but let me know if you think this needs to be explained more clearly.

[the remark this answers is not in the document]
MichaelHanksSF: Yeah, I see what you mean. I think it's clearer without they do not replace them. For here I think it's clear enough. Perhaps in the model guide we need something on this specifically?

dm_regional_app/templates/dm_regional_app/views/entry_rates.html
```diff
@@ -9,15 +9,15 @@
     <h1>Adjust entry rates</h1>
     <br>
     <p>
-        You may want to edit the number of children per year entering care to give a more accurate
+        You may want to edit the number of children per day entering care to give a more accurate
         picture of future forecasting or simply to explore what changes you may expect to see.
```
MichaelHanksSF, line 13: Good spot!

MichaelHanksSF (approved):
Works great - thanks!

PR #171, opened 2025-02-24
Optimise drop episodes

MichaelHanksSF:
No change in functionality here; just code re-write to use processes that are less computationally intensive to manage the job of:
- [ ] identifying 'redundant' episodes that represent changes in legal status/status of placement
- [ ] removing these while ensuring that contiguity of episodes is retained i.e. that we don't cut short/create gaps in care periods

### DIFF
--- a/ssda903/datacontainer.py
+++ b/ssda903/datacontainer.py
@@ -174,36 +174,32 @@ class DemandModellingDataContainer:
         combined.loc[change_ix, "DEC"] = decom_next[change_ix]
 
         # Mark episodes that represent only a change in legal status or placement status
-        combined["Skip_Episode"] = np.where(
+        combined["skip_episode"] = (
             (combined["RNE"].isin(["L", "T", "U"]))
             & (combined["CHILD"] == combined["CHILD"].shift(1))
-            & (combined["DECOM"] == combined["DEC"].shift(1)),
-            True,
-            False,
+            & (combined["DECOM"] == combined["DEC"].shift(1))
         )
 
-        # Keep episodes with Skip_Episode == FALSE, but maintain continuity in episode close info (DEC, REC, REASON_PLACE_CHANGE) with skipped episodes
-        kept_rows = []
-        skipped_episode = False
-        last_skipped_dec = None
-        last_skipped_rec = None
-        last_skipped_rpc = None
+        # Identify sequences of redundant episodes
+        # Note that this method doesn't isolate sequences of redundant episodes only; it actually increases by one whenever skip_episodes False -> True happens
+        # To identify a sequence, "skip_episode" and "redundant_group" must both be used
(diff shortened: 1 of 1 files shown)


ssda903/datacontainer.py
```diff
@@ -174,36 +174,32 @@ def combined_data(self) -> pd.DataFrame:
...
-                last_skipped_rec = row["REC"]
-                last_skipped_rpc = row["REASON_PLACE_CHANGE"]
-
-        combined = pd.DataFrame(kept_rows)
+        # Identify sequences of redundant episodes
+        # Note that this method doesn't isolate sequences of redundant episodes only; it actually increases by one whenever skip_episodes False -> True happens
+        # To identify a sequence, "skip_episode" and "redundant_group" must both be used
+        combined["redundant_group"] = (combined["skip_episode"] & ~combined["skip_episode"].shift(1, fill_value=False)).cumsum()
+
+        # Backfill the info in DEC, REC and REASON_PLACE_CHANGE from last in group to all in group
```
amynickolls, line 188: Would it never occur that the initial episode was marked as redundant, therefore having to fill DECOM?

ssda903/datacontainer.py
```diff
@@ -174,36 +174,32 @@ def combined_data(self) -> pd.DataFrame:
...
-                    skipped_episode = False
-                kept_rows.append(row)
-            else:
-                skipped_episode = True
-                last_skipped_dec = row["DEC"]
-                last_skipped_rec = row["REC"]
-                last_skipped_rpc = row["REASON_PLACE_CHANGE"]
-
-        combined = pd.DataFrame(kept_rows)
+        # Identify sequences of redundant episodes
```
amynickolls, line 183: Might be good to make clearer, between this comment and the one above, which episodes we want to keep and which we don't, because the code felt more helpful than the comments here!

amynickolls (commented):
Seems quick, I'm getting about 0.4 seconds run time on the 100k dataset. `combined.loc[mask, ["DEC", "REC", "REASON_PLACE_CHANGE"]] = combined.loc[mask_replace][["DEC", "REC", "REASON_PLACE_CHANGE"]].values` is the biggest time sink. A couple of clarifying q's on the logic as you know the 903 better than me!

ssda903/datacontainer.py
MichaelHanksSF: No, first episode can't be redundant, because otherwise we'd have no information to change DECOM to.

This line in the logic ensures the first episode is never selected:
& (combined["CHILD"] == combined["CHILD"].shift(1))

ssda903/datacontainer.py
MichaelHanksSF: Will make the comments clearer

amynickolls (approved):

PR #174, opened 2025-02-28
feat: send SSO errors to sentry

christina-moore:


amynickolls (approved):
Looks good and I've learnt something! Should we be sending all errors on the platform to sentry? Why/why not?

christina-moore:
> Looks good and I've learnt something! Should we be sending all errors on the platform to sentry? Why/why not?

Ah good question! Sentry will capture any unhandled errors automatically e.g. when users hit 500s that will be reported in sentry. For handled errors, we have a choice around what we want to do with them. It'll be an interesting one to think about, for example, in the middleware you've suggested for SE platform. We could catch the error, report it to sentry, and then also display a message to the user. I think that'll likely be a good way forward so users have feedback and also we as application monitors have feedback.

But overall I think there could be situations where it doesn't make sense to report to sentry (although logging it is always good) - kind of a case by case basis.

Issue #172, opened 2025-02-25
Bug: presence of birth-1 children in dataset cause adjusted page crash


amynickolls:
Error arise from sorting method in utils.py - rather than splitting string we should sort on index from AgeBrackets enum

PR #175, opened 2025-02-28
bugfix rate tables

MichaelHanksSF:
This PR covers a couple of different pieces of functionality:
- [ ] fixes the issue raised [here](https://github.com/SocialFinanceDigitalLabs/cs-demand-model-regional/issues/172) by using the Enum indices from AgeBrackets and PlacementCategories instead of sorting on values
- [ ] fixes the issue where the "From" model state wasn't shown on the changes made to transition rates table on the costs page
- [ ] small optimisation boost by replacing .apply() statements in rates tables with vectorised methods


dm_regional_app/utils.py
```diff
@@ -182,39 +182,48 @@ def remove_age_transitions(df):
...
     Sorts entry, exit, and transition rate tables by age and placement as
-    a standard sort_values sorts lexicographically, putting 10 befor 5.
+    a standard sort_values sorts lexicographically, putting 10 before 5.
 
     Transition rate tables need to be sorted on 'From' and 'To' in order
     to maintain sorting across tables and columns.
     """
+    # Retrieve an ordered list of the members of each relevant Enum
+    age_list = [a.value.label for a in AgeBrackets._members_by_index()]
+    place_list = [a.value.label for a in PlacementCategories._members_by_index()]
```
amynickolls, line 192: Do we need to sort on placement? And do we want this to be anything other than alphabetical? I suppose this will cover in the instance that someone adds numbers to the labels...

amynickolls (changes requested):
Looks good - just one comment on the 'Not in care' use.

dm_regional_app/utils.py
MichaelHanksSF: I think it's just good practice? I don't see why we wouldn't want to do this. If the enums do change, people may have a stronger want/need to sort here.

dm_regional_app/charts.py
```diff
@@ -433,17 +433,17 @@ def transition_rate_table(data):
     df.set_index(["from", "to"], inplace=True)
 
     # filter out children leaving care and rates for children remaining in the placement
-    df = df[df["To"].apply(lambda x: "Not in care" in x) == False]
+    df = df[~df["To"].str.contains("Not in care", na=False)]
```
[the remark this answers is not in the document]
MichaelHanksSF: I had the same thought when first reviewing, so will try to do this.

PR #178, opened 2025-03-11
Video links updated

amynickolls:
2 video links updated -

Please note I'm testing the setting on one (restricting to csdemandmodel.com only) which is why it currently isn't able to play - will check if it works on staging and update video settings accordingly.


christina-moore (approved):
Will be curious to understand how you're restricting the domain! And is it to csdemandmodel.org.uk?

PR #179, opened 2025-03-12
information tooltip added to predict end date field

amynickolls:
Information tooltip added to predict filter

### DIFF
--- a/dm_regional_app/templates/dm_regional_app/base.html
+++ b/dm_regional_app/templates/dm_regional_app/base.html
@@ -14,14 +14,7 @@
         <script src="{% static 'js/bootstrap.bundle.min.js' %}"></script>
         <script src="{% static 'js/jquery-3.7.1.min.js' %}"></script>
         <script src="{% static 'js/modal.js' %}"></script>
-        <script>
-            document.addEventListener("DOMContentLoaded", function() {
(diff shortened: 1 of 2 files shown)


dm_regional_app/templates/dm_regional_app/base.html
```diff
@@ -14,6 +14,14 @@
...
         <script src="{% static 'js/jquery-3.7.1.min.js' %}"></script>
         <script src="{% static 'js/modal.js' %}"></script>
+        <script>
+            document.addEventListener("DOMContentLoaded", function() {
+                var tooltipTriggerList = [].slice.call(document.querySelectorAll('[data-bs-toggle="tooltip"]'));
+                var tooltipList = tooltipTriggerList.map(function(tooltipTriggerEl) {
+                    return new bootstrap.Tooltip(tooltipTriggerEl);
+                });
+            });
+        </script>
```
christina-moore, line 24: I think it does make sense to initialise tooltips on a central page so that we can access them everywhere. However, I think this would best be moved into a separate script (similar to modal.js) so that we don't have JS initiliasation happening in the base template. This is more from a future-proofing mindset, if we need to write more JS we won't be cluttering the base template.

christina-moore (approved):

PR #187, opened 2025-04-16
feat: force password reset after admin password reset

christina-moore:


dm_regional_app/tests/test_scenario_detail_view.py
```diff
@@ -10,6 +10,7 @@
         "remove": [
             "dm_regional_app.middleware.scenario_middleware.SessionScenarioMiddleware",
             "dm_regional_app.middleware.force_mfa_middleware.ForceMFAMiddleware",
+            "dm_regional_app.middleware.update_password_middleware.UpdatePasswordMiddleware",
```
amynickolls, line 13: Wondering how come these tests were working before as the project settings haven't changed? Or is this just introducing good practice?

amynickolls (approved):

dm_regional_app/tests/test_scenario_detail_view.py
christina-moore: It's a good question! This one wasn't an issue as we set the force_update_password flag to False (as default) in the test builder. But yes, was thinking it would be better practice to be specific here!

PR #197, opened 2025-06-10
fix: add auth to data upload decorator

christina-moore:
The user already needed to be authenticated, but if they weren't the system was erroring out. This will give a helpful error to any end users.


amynickolls (approved):
Looks good!

PR #204, opened 2025-06-20
date conversion fix

amynickolls:
Dates now must be YYYY-MM-DD to align with data platform outputs

Samples updated to match (now also have birth-1 children)


MichaelHanksSF (approved):
Removes inf and nan rates in sample files

PR #206, opened 2025-06-23
NaNs now dropped rather than checked for

amynickolls:
Assert statements were causing failure, now NaNs are dropped instead


rtavares (dismissed):
Amy, can we have a unit test for this feature?

PR #217, opened 2025-08-12
feat: create admin group

christina-moore:
New Admin group to be used when superusers create admins - this group contains all the permissions the admins will need in order to manage regular users


amynickolls (commented):
Hello - how come there is only a migration in this PR and not any associated model code?

christina-moore:
> Hello - how come there is only a migration in this PR and not any associated model code?

Ah yes, this is a data migration rather than a schema migration. So this code will not affect the database structure at all, but it will fill in a few database rows when it's applied. We make it a migration rather than a management command or a script to be run because on every deployment we want the same database rows populated.

amynickolls (approved):
All works as expected!

PR #218, opened 2025-08-13
docs: clarify full deployment process

christina-moore:


docs/local_development.md
```diff
@@ -0,0 +1,40 @@
+# Local Development
+You need to have [python](https://www.python.org/) and [poetry](https://python-poetry).
+
+1. Create an `.env` file by copying the `.env.example` one and editing the variables as needed.
+2. Set environment variable `DJANGO_SETTINGS_MODULE=dm_regional_site.settings.development` in relevant places.
```
amynickolls, line 5: Do we want to change this to set via terminal, or do we want to do more research into why this seems to not be importing from .env first?

docs/local_development.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Load fixtures
+To load some dummy data, run:
+
+```
+python manage.py load-fixtures
+```
+
+This will load some users:
+
```
amynickolls, line 19: More information here about what happens, any prior data in the database will be destroyed, and the below users will be populated in the database, allowing sign in and LA to be added too.

docs/local_development.md
```diff
@@ -0,0 +1,40 @@
+# Local Development
+You need to have [python](https://www.python.org/) and [poetry](https://python-poetry).
+
+1. Create an `.env` file by copying the `.env.example` one and editing the variables as needed.
+2. Set environment variable `DJANGO_SETTINGS_MODULE=dm_regional_site.settings.development` in relevant places. 
+3. Install dependencies with `poetry install`; 
+4. Launch a poetry shell so the virtual environment is active: `poetry shell`;
```
amynickolls, line 7: We discussed changing poetry instructions in an earlier meeting. Shell becoming outdated.

docs/local_development.md
christina-moore: I've added the database flush and users can be used to log in. I think any further instructions should be referenced in a user guide as that's how to start using the tool 🙂

docs/architecture.md
```diff
@@ -1,71 +1,4 @@
...
-        Rel(data_storage, child_level_data, "")
-
-    }
-
-}
-
-UpdateLayoutConfig($c4ShapeInRow="3", $c4BoundaryInRow="1")
-
-
+![img.png](../docs/images/architecture.png)
```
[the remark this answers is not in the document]
christina-moore: Nice spot!

rtavares:
It would be nice to have a PR description.

PR #222, opened 2025-09-08
feat: add email verification for mfa

christina-moore:
It's best practice to have users verify their emails before setting up MFA. This PR adds the capability to send emails via sendgrid and then the django-allauth implementation of email verification via code. There are limitations of this implementation from the package with the largest being:
- When a code expires, there is no straightforward way to tell the user that. Code expiration is implemented in a very nested module in django-allauth and trying to surface it would add complexity that is harder to maintain and more brittle to any package updates. If a code expires, the user will be prompted to login again and will be sent a new verification code - this shouldn't be too vulnerable to spamming messages as the expiration is set at 10 minutes. I've tried to work around this by alerting the user in the verification page that there is an expiration, but happy to take ideas on alternate implementations!

image

image

image


amynickolls (commented):
It works - do we want to implement this so that it doesn't apply in development?

christina-moore:
> It works - do we want to implement this so that it doesn't apply in development?

Yes good shout - that's moved to prod settings now https://github.com/SocialFinanceDigitalLabs/cs-demand-model-regional/pull/222/commits/533ef969d23292d52d7c66907907615e22fc4f74

amynickolls (dismissed):
Looks good!

amynickolls (approved):

PR #223, opened 2025-09-09
feat: remove user permissions from admin

christina-moore:
Removes the ability for admin users to adjust permissions on regular users within the admin panel.
Note: this does not change their abilities to do so programmatically!

image


amynickolls (approved):
Looks good! Hopefully admins don't go round unassigning their permissions :D

PR #224, opened 2025-09-09
feat: set session cookie expire default to 5 days

christina-moore:
Django defaults to allowing sessions to stay alive for 14 days. The penetration test results dictated that that was too long. We need to find the middle ground between usability and security! So have defaulted to 5 days for now, but it can be configured as an env var in Heroku

### DIFF
--- a/dm_regional_site/settings/base.py
+++ b/dm_regional_site/settings/base.py
@@ -173,6 +173,8 @@ AUTH_PASSWORD_VALIDATORS = [
     },
 ]
 
+SESSION_COOKIE_AGE = config("SESSION_COOKIE_AGE", 432000)
+
 
 # Internationalization
 # https://docs.djangoproject.com/en/4.2/topics/i18n/


amynickolls:
Hello - want to understand something of the requirements here. I think our 'session scenarios' operate outside of a Django cookie but are conceptually similar. What's the expected outcome with this change, if purely about authentication (a user gets logged out after 5 days) then this is fine, but if we are thinking just about what information we are assigning to a user, do we need to do something with the session scenarios too?

christina-moore:
> Hello - want to understand something of the requirements here. I think our 'session scenarios' operate outside of a Django cookie but are conceptually similar. What's the expected outcome with this change, if purely about authentication (a user gets logged out after 5 days) then this is fine, but if we are thinking just about what information we are assigning to a user, do we need to do something with the session scenarios too?

Very good question! So this setting will expire the whole session including the scenario id as we are storing that within Django's sessions. What are your thoughts on usability here and how long we want the scenarios to last? (Currently they're being reset when the session expires at 2 weeks)

amynickolls:
> Very good question! So this setting will expire the whole session including the scenario id as we are storing that within Django's sessions. What are your thoughts on usability here and how long we want the scenarios to last? (Currently they're being reset when the session expires at 2 weeks)

I think that's fine, our initial thinking was that sessions would be wiped once a week anyway. Potentially we should think about adding to our user docs - I can have a look at that. But then it's possible that having the option to save scenarios could be enough of a prompt for users to think that their inputs might not be kept over multiple days!

Unsure if there's a way of testing this other than waiting for 5 days? Or if this is good to approve based on my understanding of Django docs?

PR #227, opened 2025-09-19
Remove header fields

MichaelHanksSF:
Allows header files without MOTHER, MC_DOB and UPN to be recognised as header files. Files with these fields will also still be recognised.


amynickolls (commented):
For completeness, do we not want to remove any field we aren't using from each header/episodes/UASC? Otherwise we might end up needing to come back to this with another change.

I guess there's also a question of whether we keep the schemas of the tables we aren't using.

amynickolls (approved):
I'm in agreement with this!

PR #267, opened 2026-03-17
UI updates

amynickolls:
This makes 3 changes:

- REGION and REGIONAL DATA SHARING AGREEMENT on homepage changed to pass env variable so it's customisable by region
- typo fix - accomodation to accommodation
- sort unique ethnicity and LA so that these options appear alphabetically in the historic filtering dropdown


MichaelHanksSF (changes requested):
Could you add the new env variables to the env.example file? Otherwise all good!

MichaelHanksSF (approved):
Tested and approved

PR #268, opened 2026-03-17
remove int typing

amynickolls:
Turns out, plotly will display a round number float as a whole number anyway, so casting as int was not necessary here. The `.astype(int)` was the part causing the problem - `.round()` will ignore NANs or infs


MichaelHanksSF (approved):
Approved

PR #270, opened 2026-03-19
costs chart fix and rounding

amynickolls:
Fixes an issue where with low populations (and so fewer transitions) population doesn't fill to the data end date.


MichaelHanksSF:
Can you please provide me with an example or more context on what this is trying to fix, as I can't re-create the issue that you're referring to.

MichaelHanksSF (approved):
Tested and approved

PR #272, opened 2026-03-27
remove int casting

amynickolls:
int casting is unnecessary here as plotly will display whole floats as ints anyway, plus type casting will crash with the presence of NANs, which `round()` will not


MichaelHanksSF (approved):
Tested and approved

PR #273, opened 2026-03-27
update python to 3.13

amynickolls:
This PR makes the following changes to upgrade python:

- brings `populate_same_state_transition`, `build_transition_rates_matrix`, `fill_missing_states` and `BaseModelPredictor` into the repo to enable removal of demand-model cli tool package which is not compatible with required pandas upgrade. assuming this is preferable to having to also maintain cli tool
- charts changes due to new pandas not liking my old solution to make historic and prediction lines meet
- upgrade to `ffill` and `bfill` methods
- getting rid of inplace options where they no longer work

~Also fixes bug where filtering 'Exclude UASC' returned no children - this does exist on main so isn't due to current upgrade, must've become an issue with a different package upgrade.~

I've tried to find as many bugs as possible but this needs a thorough review of functionality


pyproject.toml
```diff
@@ -21,14 +20,14 @@ django-select2 = "^8.4.1"
 django-tables2 = "^2.7.0"
 django-filter = "^24.3"
 whitenoise = "^6.6.0"
-psycopg2 = "^2.9.9"
```
christina-moore, line 24: Hmm why is this removed? We'll need it in staging/prod to connect to the database

christina-moore (commented):
~~Just for future reference, I think it's best to keep something like a version upgrade in a separate PR so changes required for that don't get confused with features/fixes - it makes it easier to review and also roll back if necessary~~
Just properly read your comment - wow that these are all for the upgrade! Big one this time 😅

MichaelHanksSF (approved):
Tested and approved

PR #275, opened 2026-03-30
fix Exclude UASC filter

amynickolls:
Exclude UASC filter not working in main - this fixes that


MichaelHanksSF (approved):
Tested and approved

PR #276, opened 2026-04-02
additional logic when loading scenarios

amynickolls:
currently loading an old scenario might mean that date parameters are outside of data start/end dates, this fixes that. this is important on account of the odd behaviour SE were having when the reference start date was before when the data start date should have been.

also adds a warning for users if the data has been updated between the last time the scenario was updated and now - this lets users know that outputs may look different.

other data changes shouldn't break the model due to the way we filter


MichaelHanksSF (approved):
Tested and approved

PR #277, opened 2026-04-02
restrict prediction end date

amynickolls:
as demonstrated by Patrick in our stress testing session, users trying to create an extremely long prediction could cause a timeout on the page. this prevents that behaviour


MichaelHanksSF (approved):
Tested and approved

PR #283, opened 2026-05-19
refactor and fix raw_transition_rates

MichaelHanksSF:
This PR does a few things:
- [ ] removes backfill solution to shifted stock; previously we used backfill to impute a stock value for day 0, as we were truncating stock prior to shifting and aligning with transitions. This solution now i) shifts stock prior to shifting so we have a real value for day 0; ii) does not backfill for day -1 i.e. where there is no data to supply a real value; instead this will return a value of NaN for stock and therefore also for daily rate i.e. first day of data is now ignored for rate calculation
- [ ] implicitly ignores NaN daily rate values for average rate calculation; previously we use fillna(0) before calculating mean rates; this is an error as it means that 0s that are created by 0/0 (where there was no stock and therefore transitions were impossible) count the same as 0s created by 0/N (where there was stock and transitions were possible)
- [ ] raises valueerror if inf is created - this should be impossible I think in the current setup as stock and transitions are worked out from the same logic; so something should be seriously wrong if this happens
- [ ] refactors the function so that it can be more easily unit tested
- [ ] adds unit tests for the function


ssda903/population_stats.py
```diff
@@ -7,6 +7,69 @@
...
+    # Catch issues where shifted stock = 0 but transitions is a positive number
+    # This should be impossible, but will cause the model to break
+    inf_transitions = (transitions_trunc.gt(0) & stock_trunc.eq(0))
+    if inf_transitions.any().any():
+        inf_locs = inf_transitions.stack()
+        inf_locs = inf_locs[inf_locs].index
+        sample_locs = list(inf_locs[:10])
+        raise ValueError(
+            "Invalid calculation: found transitions > 0 where previous stock == 0. " \
+            f"Count = {len(inf_locs)}. Sample locations = {sample_locs}"
```
amynickolls, line 55: I'm also going to add an issue to handle ValueErrors within the frontend, as I think it's "possible" for this to happen when the model is filtered, even if it doesn't happen when validating the file on upload. I think we will want to serve this error direct to the user.

amynickolls (dismissed):

amynickolls (approved):

PR #291, opened 2026-06-12
Age refactor

MichaelHanksSF:
This PR refactors the age_age_change_episodes function:
- [ ] removes all lambda operations
- [ ] refactors to avoid repetition of operations
- [ ] refactors to improve testability
- [ ] adds unit tests

The actual operation of the function is the same and produces identical results using sample data than previous code.


ssda903/datacontainer.py
```diff
@@ -333,70 +443,35 @@ def _add_age_change_eps(self, combined: pd.DataFrame) -> pd.DataFrame:
...
-            # Assign the index of the age_brackets_df relevant to each row
-            df["bracket_index"] = pd.cut(
-                df[age_col], bins=age_bins, labels=age_brackets_df.index, right=False
-            )
-            # Map the attribute to the index
-            df[return_col] = age_brackets_df.loc[df["bracket_index"], attribute].values
-
-            return df.drop(columns=["bracket_index"])
+        age_bins = np.sort(np.unique(age_brackets_df[["start", "end"]].values.ravel()))
+        age_bounds = np.array([b.value.end for b in AgeBrackets if b.value.end])
```
amynickolls, line 453: some comments here on what the outputs of these are and why we need both would be helpful

ssda903/datacontainer.py
```diff
@@ -405,43 +480,52 @@ def get_age_bracket_attribute(
         # RNE = Reason for new episode
```
amynickolls: this comment is in the wrong place now

amynickolls (commented):
Some really minor comments around comments - this is great! 4x faster

amynickolls (approved):

PR #295, opened 2026-07-03
build: update dependabot action

christina-moore:


Lady-Frost (approved):
Dependabot config checks for slack notifications look all good ✅

PR #296, opened 2026-07-17
build: require https and secure session cookies

christina-moore:
Off the back of the penetration test, we need to add some django settings in production to:
- Ensure requests can only come through https and any http requests are automatically forwarded
- Add "secure" flag to session cookie to ensure it is not shown in plaintext when visiting any http urls in its scope
- Add "secure" flag to CSRF cookie
- Prevent session cookie from being sent to another domain

Docs: https://docs.djangoproject.com/en/6.0/ref/settings


Lady-Frost (approved):
Security settings check out! 

Session must be protected with Strict and CSRF

All good in my eyes ✅

PR #297, opened 2026-07-17
feat: ensure redirect urls are from the site

christina-moore:
If an attacker tries to redirect maliciously (e.g. for a phishing attempt) to another site, this will no longer be allowed.

Question - do we want the default redirect to home (as in the code currently) or elsewhere?


amynickolls:
I think a default redirect is fine, but we should add a message to users if they are landing on home due to a link not being safe. Do we want to also log an error to notify us too?

For the user, maybe "You have been returned to the homepage due to a broken link, please contact your administrator to resolve."?

Lady-Frost:
I believe we have 3 options here 

1. Redirect to a default set page (home or a portal or any landing page within the application which is always available to the user, home pages are great for this) 
I would say this is the path of least resistance, only worth considering which page.

2. Redirect them to an error page. This is rather old school and not great for a user journey because all you can do on that page is perceive the error, like a 404 you 
page is not found or in this case I suppose a 503 forbidden. While I think these error pages are great to illustrate breakage and the fact that something truly is wrong,
in this case it is solely for the attacker and a regular user would not benefit from that status. If anything they are now stuck in a more or less dead end.
Further you now need to explicitly navigate to menu and where you need to go, adding a few more clicks. 
Unless we think there is value in communicating that to the users I would not advise this option

3. Pop ups! Provide the user with a pop up overly the moment the malicious url of different domain origin is detected. This still prevents the flow of attacker redirection 
but does provide the option to the user to pause, potentially be presented with the destination URL of where they are being redirected and in case they do in fact want to continue 
down that path, they can - OR - the click/navigate towards where we were about to redirect them for safety.
This gives the user a choice but is more work to develop the popup and interface

About whether the user should be informed, assuming it is not an error page. This needs to happen upon redirecting. In other words: 

Malicious Navigation -> Defensive Redirection -> Loading the default redirection page -> presenting the user with that information 

So wherever they are landing to must support the functionality to present an optional message whether it is a toastr or banner UI, either way a temperary message,
which fades after a timeout. 
Personally I don't think this is worth it (please do challenge my biase opinion) because it is around the same effort ish compared to point #2.
Also I would not inform them of a broken link because that tailors their perception that a link is broken and moves fault onto us/the admin? 
Gracefully redirecting users away from external links should not be something they are surprised about in the user journey. 
But you folks know the user journey much better than what I am assuming in this moment 

Lastly, we should be logging this as a "warning" somewhere. Just basic user info telemetry + where they were trying to be redirected to, 

Much love 💙

christina-moore:
Thanks @amynickolls and @Lady-Frost for your thoughts! I think logging the error is a great idea. In terms of displaying a message to the user that is actually relatively easy through django as banner messages are builtin. But I think I disagree with the need for a user message here a - I think we can assume only a malicious user would tamper directly with the URL such that they are trying to direct to an external website. And I don't think extra info is needed for malicious users

Lady-Frost:
@christina-moore  Makes sense! 

Oh banners in Django, forgot about those, excellent point actually!

I also don't think the additional information would be of benefit to users here, I would assume this case is for the maliciously acting ones and then if for whatever reason we get a significant amount of User Journey issues from the normal user pool then we can revisit this again. 

As always keeping things as simple as possible is a solid course of action 

This makes total sense to me

Lady-Frost (approved):
Redirect checks out 

The control over the environment variable for the hosts should give us sufficient flexibility to cases 

All good ✅

